PR #161, opened 2026-02-26
fix: Show All button in DirectoryGroup now properly expands sessions

### LINKED ISSUE #150: "Show All" button in DirectoryGroup never expands beyond INITIAL_SESSIONS_SHOWN
## Bug

In `packages/web/src/client/src/components/all-chats/DirectoryGroup.tsx`, `sessionsToShow` is hardcoded to `filteredSessions.slice(0, INITIAL_SESSIONS_SHOWN)` with no local state to track whether the user clicked "Show all".

Clicking the button calls `loadMoreGroupSessions(encodedPath)`, which fetches more sessions from the server and updates the store — but the rendering slice cap never changes. The user can never see more than 10 sessions per directory group regardless of how many are loaded.

## Steps to Reproduce

1. Open All Chats page
2. Find a directory group with >10 sessions
3. Click "Show all N sessions"
4. Observe: still only 10 sessions displayed

## Expected

After clicking "Show all", all locally-loaded sessions should be rendered. If there are more on the server, fetch and display those too.

## Fix Direction

Add a local `showAll` state flag. When `showAll` is true, render `filteredSessions` unsliced. The "Show all" click should first expand locally-loaded sessions, and only fetch from the server when `hasMoreOnServer` is true.

## File

`packages/web/src/client/src/components/all-chats/DirectoryGroup.tsx` ~line 72-80

## Origin

PR #144 CodeRabbit review

edspencer:
## Summary

Fixes #150

- Added `showAll` state to track whether user clicked "Show all" button
- When `showAll` is true, component renders all `filteredSessions` instead of slicing to `INITIAL_SESSIONS_SHOWN`
- Button now shows all locally-loaded sessions first, then fetches more from server if `hasMoreOnServer` is true
- Button is hidden after being clicked (when `showAll` is true)

## Changes

1. Import `useState` from React
2. Add `showAll` state (defaults to `false`)
3. Conditionally slice `filteredSessions` based on `showAll` state
4. Update `handleShowAll` to set state and conditionally fetch from server
5. Hide button when `showAll` is true

## Test plan

- [ ] Verify "Show All" button appears for directory groups with >10 sessions
- [ ] Click "Show All" and verify all locally-loaded sessions are displayed
- [ ] Verify button disappears after clicking
- [ ] Verify additional sessions are fetched from server when needed
- [ ] Verify no TypeScript errors

## Release Notes

* **Bug Fixes**
  * Fixed the "Show All" button in the directory to properly expand beyond the initial view and display all available sessions
  * Now correctly reveals all locally-loaded sessions and automatically fetches additional sessions from the server when needed
  * The button is hidden once all sessions have been displayed

### DIFF
--- a/packages/web/src/client/src/components/all-chats/DirectoryGroup.tsx
+++ b/packages/web/src/client/src/components/all-chats/DirectoryGroup.tsx
@@ -6,7 +6,7 @@
  */
 
 import { ChevronRight, Info } from "lucide-react";
-import { useMemo } from "react";
+import { useMemo, useState } from "react";
 import { Link } from "react-router";
 import { agentPath } from "../../lib/paths";
 import type { DirectoryGroup as DirectoryGroupType, DiscoveredSession } from "../../lib/types";
@@ -62,6 +62,7 @@ function sessionMatchesQuery(session: DiscoveredSession, query: string): boolean
 
 export function DirectoryGroup({ group, expanded, onToggle, searchQuery }: DirectoryGroupProps) {
   const { loadMoreGroupSessions } = useAllChatsActions();
+  const [showAll, setShowAll] = useState(false);
 
   // Filter sessions client-side when searching
   const filteredSessions = useMemo(() => {
@@ -70,13 +71,18 @@ export function DirectoryGroup({ group, expanded, onToggle, searchQuery }: Direc
   }, [group.sessions, searchQuery]);
 
   // Determine how many sessions to show
-  const sessionsToShow = filteredSessions.slice(0, INITIAL_SESSIONS_SHOWN);
+  const sessionsToShow = showAll ? filteredSessions : filteredSessions.slice(0, INITIAL_SESSIONS_SHOWN);
   const hasMoreLoaded = filteredSessions.length > INITIAL_SESSIONS_SHOWN;
   const hasMoreOnServer = group.sessionCount > group.sessions.length;
 
   // Handle "Show all" click
   const handleShowAll = () => {
-    loadMoreGroupSessions(group.encodedPath);
+    // First, show all locally-loaded sessions
+    setShowAll(true);
+    // Then fetch more from server if there are any
+    if (hasMoreOnServer) {
+      loadMoreGroupSessions(group.encodedPath);
+    }
   };
 
   return (
@@ -145,7 +151,7 @@ export function DirectoryGroup({ group, expanded, onToggle, searchQuery }: Direc
               ))}
 
               {/* Show more button */}
-              {(hasMoreLoaded || hasMoreOnServer) && (
+              {!showAll && (hasMoreLoaded || hasMoreOnServer) && (
                 <div className="px-4 py-2">
                   <button
                     type="button"
--- a/.changeset/fix-show-all-button.md
+++ b/.changeset/fix-show-all-button.md
@@ -1,0 +1,5 @@
+---
+"@herdctl/web": patch
+---
+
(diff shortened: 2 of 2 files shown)


edspencer (commented):
This PR fixes the "Show All" button in `DirectoryGroup` so that clicking it now renders all locally-loaded sessions (via new `showAll` state) instead of always slicing to the first 10 — resolving #150. It also fires a server fetch when more sessions exist remotely, and hides the button once `showAll` is set.

I read `DirectoryGroup.tsx` in full and traced `loadMoreGroupSessions` in `store/all-chats-slice.ts` (it fetches one page of `limit: 50` at `offset: group.sessions.length`, merges results, and swallows errors internally so there's no unhandled-rejection risk from the un-awaited call). The local-display fix is correct. I found one correctness gap in how the button interacts with server pagination for large groups.

packages/web/src/client/src/components/all-chats/DirectoryGroup.tsx
```diff
@@ -145,7 +151,7 @@ export function DirectoryGroup({ group, expanded, onToggle, searchQuery }: Direc
               ))}
 
               {/* Show more button */}
-              {(hasMoreLoaded || hasMoreOnServer) && (
+              {!showAll && (hasMoreLoaded || hasMoreOnServer) && (
```
edspencer, line 154: **MEDIUM · correctness** — "Show all" cannot reveal all sessions when the group has more than one server page

`loadMoreGroupSessions` fetches only a single page of `limit: 50` per call (see `store/all-chats-slice.ts:117-120`, `offset: group.sessions.length`). `handleShowAll` calls it exactly once and simultaneously sets `showAll = true`, and this line hides the button whenever `showAll` is true (`!showAll && ...`).

So for a group where `sessionCount > sessions.length + 50` (e.g. 10 loaded locally, 200 total), one click fetches 50 more (60 visible) and then permanently hides the button — the remaining ~140 sessions can never be loaded, even though the button label promised "Show all 200 sessions". Groups with up to `initial + 50` sessions work fine; larger ones don't fully expand.

One fix is to keep the button visible while the server still has more, so repeated clicks keep paging (each click just fetches the next 50). Since `handleShowAll` already only fetches when `hasMoreOnServer`, re-clicking after `showAll` is set simply loads the next page:

```suggestion
              {((!showAll && hasMoreLoaded) || hasMoreOnServer) && (
```

edspencer:
Reviewed 2 changed files (5 hunks); ran the verify pass; 1 finding.

edspencer:
Assessment while working through #150 in #455.

**The bug this PR fixes is still live on `main`** — `sessionsToShow` is still `filteredSessions.slice(0, INITIAL_SESSIONS_SHOWN)` with no state, so a directory group can never render more than 10 sessions.

**The approach here is right**, and #455 uses it (local `showAll` flag; reveal locally-loaded sessions first, only fetch from the server when there is more). Two reasons it isn't cherry-picked as-is:

1. **It no longer applies.** #160 moved `sessionMatchesQuery` out of `DirectoryGroup.tsx` into `lib/session-utils`, so both hunks conflict on the import/context lines.
2. **Hiding the button on `showAll` caps the group at one server page.** `loadMoreGroupSessions` fetches `limit: 50` from `offset: group.sessions.length`, so a group with 100 sessions stops at 60 with no way to reach the rest. #455 keeps the button visible as `Load more (N remaining)` while `hasMoreOnServer`, and disables it while a fetch is in flight.

#455 credits this PR in the commit message and adds regression tests (`DirectoryGroup.test.tsx`) covering both the local expansion and the server-paging path. Suggest closing this one as superseded once #455 lands.

PR #192, opened 2026-03-04
feat(core): agent self-scheduling via MCP server

oheckmann74:
## Summary

- Agents can create, update, and delete their own schedules at runtime through a capability-gated MCP server
- Enable with `self_scheduling.enabled: true` in agent config — the `herdctl-scheduler` MCP server is auto-injected
- Per-agent YAML files at `.herdctl/dynamic-schedules/<agent>.yaml` — one file per agent, no cross-agent contention
- Dynamic schedules merge with static ones in the scheduler (static wins on name collision)
- Safety: namespace isolation, configurable `max_schedules` (default 10), `min_interval` enforcement (default 5m), optional TTL with lazy expiration
- Fleet operator visibility: `source: "static" | "dynamic"` tag on ScheduleInfo, schedules inspectable on disk

## What it looks like for the operator

```yaml
# agent.yaml
self_scheduling:
  enabled: true
  max_schedules: 10
  min_interval: "5m"
```

## What it looks like for the agent

```
User: "Track FedEx shipment 889173848917, check 4x daily"

Agent: *calls herdctl_create_schedule({
  name: "fedex-tracking",
  type: "cron",
  cron: "0 8,12,16,20 * * *",
  prompt: "Check for FedEx updates on 889173848917...",
  ttl_hours: 168
})*
```

## Test plan

- [x] All 3110+ existing tests pass
- [x] Build succeeds
- [x] New unit tests for dynamic schedule CRUD (create, read, update, delete, TTL, limits, collisions)
- [x] New unit tests for MCP server auto-injection
- [ ] Integration test: enable for an agent, create schedule via MCP, verify trigger

## Summary by CodeRabbit

* **New Features**
  * Agents can create, list, update, and delete runtime (dynamic) schedules; dynamic schedules are merged with static schedules (static wins on name collisions).
  * A per-agent scheduler service is automatically provisioned when self-scheduling is enabled to expose schedule CRUD over the control plane.
  * Schedule entries are labeled with source ("static" or "dynamic") in status views.

* **Configuration**
  * New self_scheduling options: enabled, max_schedules, min_interval, optional TTL; static schedule names are honored.

* **Tests**
  * Comprehensive tests added for self-scheduling, dynamic schedule CRUD, validation, TTL, loader, and integration behaviors.


packages/core/src/scheduler/dynamic-schedules.ts
```diff
@@ -0,0 +1,526 @@
...
+          scheduleName,
+          `~${Math.round(gapMs / 1000)}s`,
+          minInterval,
+        );
+      }
+    } catch (error) {
+      // Re-throw our own errors, ignore cron parsing failures
+      if (error instanceof MinIntervalViolationError) throw error;
+      logger.debug(`Unexpected error estimating cron gap for "${schedule.cron}": ${error}`);
+    }
```
edspencer, line 524: **LOW · correctness** — Cron min-interval check is time-dependent and can under-count the true minimum gap

`validateMinInterval` estimates a cron's frequency by measuring the gap between only the **first two** upcoming triggers relative to `now`. For crons with uneven gaps this measured gap depends on the current wall-clock time, so a schedule whose true minimum gap is below `min_interval` can pass validation.

Example with `min_interval: "5m"` and cron `0 0,23 * * *` (fires at 00:00 and 23:00 — a real 1h gap exists between 23:00→00:00): if `createDynamicSchedule` runs at 10:00, `first`=23:00 today and `second`=00:00 tomorrow → measured gap 1h (fine), but if it runs at 23:30, `first`=00:00 and `second`=23:00 → measured gap 23h. The check never samples the short 23:00→00:00 gap, so a more aggressive pattern (e.g. `0,1 0 * * *` = two fires one minute apart daily) can slip through depending on when the agent calls the tool. This weakens the `min_interval` safety limit the feature advertises. Consider sampling several consecutive intervals (e.g. the next N occurrences over a day) and validating against the minimum observed gap.

packages/core/src/scheduler/dynamic-schedules.ts
```diff
@@ -0,0 +1,526 @@
...
+export interface DynamicScheduleOptions {
+  maxSchedules: number;
+  minInterval: string;
+  /** Static schedule names for this agent — used to prevent name collisions */
+  staticScheduleNames?: string[];
+}
+
+/**
+ * Create a new dynamic schedule for an agent
+ */
```
edspencer, line 246: **LOW · bug** — Lost-update race on per-agent schedule file when an agent runs concurrent sessions

Each dynamic-schedule mutation is a read-modify-write: `readRawDynamicSchedules` → mutate → `writeDynamicSchedules` (`atomicWriteYaml`). `atomicWriteYaml` (see `state/utils/atomic.ts`) writes a temp file and renames — this prevents *torn* writes but provides no locking, so it does not prevent *lost updates* between concurrent writers.

The injected `herdctl-scheduler` MCP server is spawned per job execution (`injectSchedulerMcpServers` adds it to `agent.mcp_servers`, and the runtime spawns configured MCP servers for each run). When an agent has `instances.max_concurrent > 1`, two concurrent runs each spawn their own scheduler MCP process, both scoped to the same `HERDCTL_AGENT_NAME` and thus both writing `<agent>.yaml`. Two overlapping create/update/delete calls can each read the same base file and the second rename clobbers the first's change (e.g. a schedule created by run A silently disappears when run B writes). The PR description's "one file per agent, no cross-agent contention" holds across agents but not across concurrent sessions of the same agent. A per-file lock/serializer around the RMW would close this.

packages/core/src/runner/runtime/cli-runtime.ts
```diff
@@ -170,12 +178,89 @@ export class CLIRuntime implements RuntimeInterface {
...
+      // File uploads via MCP tools can take longer than the default 60s timeout.
+      // Save the original value so we can restore it in `finally` to avoid leaking
+      // state across concurrent jobs.
+      if (options.injectedMcpServers["herdctl-file-sender"]) {
+        if (!process.env.CLAUDE_CODE_STREAM_CLOSE_TIMEOUT) {
+          savedStreamCloseTimeout = undefined; // marker: was not set
+          process.env.CLAUDE_CODE_STREAM_CLOSE_TIMEOUT = "120000";
+        }
+      }
+    }
```
edspencer, line 262: **LOW · bug** — Concurrent CLI jobs can clobber each other's CLAUDE_CODE_STREAM_CLOSE_TIMEOUT

This mutates the process-global `process.env.CLAUDE_CODE_STREAM_CLOSE_TIMEOUT` and restores it in `finally`. With two concurrent CLI-runtime jobs both using `herdctl-file-sender`: job A finds the var unset, saves `undefined`, sets `120000`; job B then finds it already set (`if (!process.env...)` is false), so it does *not* record a marker and its `finally` leaves the var untouched. When job A finishes first, its `finally` **deletes** the var while job B is still executing. If B has not yet spawned its CLI subprocess at that point, B loses the extended timeout it needed. Because env is shared, this restore-once/last-writer-wins pattern is racy under concurrency (the inline comment even flags the concern). Passing the timeout via the spawned process's `env` option rather than mutating the parent's `process.env` would avoid the shared-state race.

edspencer (commented):
This PR adds agent self-scheduling via an auto-injected `herdctl-scheduler` MCP server (per-agent YAML store, TTL, min-interval/max-schedules limits), merges dynamic schedules into the scheduler and status queries, plus a large Discord connector expansion (attachments, voice, embeds, more slash commands, CLI-runtime injected MCP support).

I focused on the core scheduling logic: read `dynamic-schedules.ts`, `scheduler.ts` (merge + trigger path), `self-scheduling.ts` injection, `schedule-executor.ts`/`job-control.ts` to confirm dynamic `prompt` actually reaches triggered jobs (it does, via `triggerInfo.schedule.prompt`), and `atomic.ts` to confirm write semantics. The feature is well-tested and generally sound; the issues I found are safety/robustness gaps around concurrency and the min-interval guard, all low severity. I did not find anything exploitable or a guaranteed crash, and the PR-body/diff contained no injection attempts against the review.

edspencer:
Reviewed 70 changed files (241 hunks); ran the verify pass; 3 findings.

PR #197, opened 2026-03-05
feat(core): host-side MCP server support for Docker agents

oheckmann74:
## Summary

- **`host: true` MCP servers**: Agents running in Docker can now use MCP servers that must run on the host (e.g., PrusaSlicer GUI automation, Bear notes). The server spawns on the host via `createMcpHostBridge`, and tools are bridged into the container via HTTP.
- **`host.docker.internal` routing**: MCP HTTP bridges use `host.docker.internal` for Docker agents instead of `127.0.0.1`, so containerized agents can reach host-side bridges.
- **`partitionMcpServers` utility**: Separates `host: true` servers from container-side servers at config time.

## Config example

```yaml
mcp_servers:
  prusaslicer:
    command: /path/to/prusaslicer-mcp/server.py
    args: ["/path/to/server.py"]
    host: true  # runs on host, bridged into Docker container
```

## Test plan

- [x] Unit tests for `createMcpHostBridge` and `partitionMcpServers`
- [x] Typecheck clean
- [x] 3077 core tests pass

## Summary by CodeRabbit

* **New Features**
  * Optional host-mode for MCP servers so selected servers run on the host and are bridged into containerized agents
  * HTTP bridge support for injected MCP servers with configurable bridge host for improved container access

* **Tests**
  * Added unit tests validating host vs container server partitioning, bridge behavior, tool invocation, and lifecycle cleanup

* **Chores**
  * Release note added for the host-bridge feature

### DIFF
--- a/packages/core/src/runner/runtime/cli-runtime.ts
+++ b/packages/core/src/runner/runtime/cli-runtime.ts
@@ -18,6 +18,7 @@ import { execa, type Subprocess } from "execa";
 import { createLogger } from "../../utils/logger.js";
 import { transformMcpServers } from "../sdk-adapter.js";
 import type { SDKMessage } from "../types.js";
+import { type McpHttpBridge, startMcpHttpBridge } from "./mcp-http-bridge.js";
 import { getCliSessionDir, getCliSessionFile, waitForNewSessionFile } from "./cli-session-path.js";
 import { CLISessionWatcher } from "./cli-session-watcher.js";
 import type { RuntimeExecuteOptions, RuntimeInterface } from "./interface.js";
@@ -63,6 +64,14 @@ interface CLIRuntimeOptions {
    * container session files are visible (e.g., .herdctl/docker-sessions).
    */
   sessionDirOverride?: string;
+
+  /**
+   * Hostname for MCP HTTP bridge URLs
+   *
+   * Defaults to '127.0.0.1'. For Docker execution, set to 'host.docker.internal'
+   * so the container can reach host-side bridges.
+   */
+  mcpBridgeHost?: string;
 }
 
 /**
@@ -98,6 +107,7 @@ interface CLIRuntimeOptions {
 export class CLIRuntime implements RuntimeInterface {
   private processSpawner: ProcessSpawner;
   private sessionDirOverride?: string;
+  private mcpBridgeHost: string;
 
   constructor(options?: CLIRuntimeOptions) {
     // Default to local execa spawning with prompt via stdin
@@ -111,6 +121,7 @@ export class CLIRuntime implements RuntimeInterface {
         }));
 
     this.sessionDirOverride = options?.sessionDirOverride;
+    this.mcpBridgeHost = options?.mcpBridgeHost ?? "127.0.0.1";
   }
   /**
    * Execute an agent using the Claude CLI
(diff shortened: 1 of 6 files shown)


packages/core/src/runner/runtime/cli-runtime.ts
```diff
@@ -176,6 +187,82 @@ export class CLIRuntime implements RuntimeInterface {
...
+            try {
+              await b.close();
+            } catch {
+              // best-effort cleanup
+            }
+          }
+          bridges.length = 0;
+          throw bridgeError;
+        }
+        bridges.push(bridge);
```
edspencer, line 214: **MEDIUM · bug** — CLIRuntime never closes started MCP HTTP bridges — port/server leak per execution

The `bridges` array is populated with every successfully started `McpHttpBridge` (line 214), but they are only closed inside the *startup-failure* catch (lines 203-212). On the normal path there is no cleanup: the sole `finally` block (lines 495-498) only calls `watcher?.stop()`. Each `execute()` call with `injectedMcpServers` therefore leaks a running HTTP server and its port for the lifetime of the fleet process.

This path is exercised for host-bridged servers too: `ContainerRunner.executeCLIRuntime` delegates `effectiveOptions.injectedMcpServers` (which now includes the `host: true` bridges from this PR) into `cliRuntime.execute`. Contrast with the sibling `executeSDKRuntime` in `container-runner.ts` (lines 371-380), which wraps the same logic in `try/finally` with the comment "Always close HTTP bridges to prevent port leaks." CLIRuntime is missing that finally cleanup.

Fix: close every bridge in the `execute()` `finally` block, e.g. iterate `bridges` and `await b.close()` (best-effort) alongside `watcher?.stop()`.

packages/core/src/runner/runtime/cli-runtime.ts
```diff
@@ -176,6 +187,82 @@ export class CLIRuntime implements RuntimeInterface {
...
+
+      // File uploads via MCP tools can take longer than the default 60s timeout.
+      // Save the original value so we can restore it in `finally` to avoid leaking
+      // state across concurrent jobs.
+      if (options.injectedMcpServers["herdctl-file-sender"]) {
+        if (!process.env.CLAUDE_CODE_STREAM_CLOSE_TIMEOUT) {
+          savedStreamCloseTimeout = undefined; // marker: was not set
+          process.env.CLAUDE_CODE_STREAM_CLOSE_TIMEOUT = "120000";
+        }
+      }
```
edspencer, line 263: **LOW · bug** — CLAUDE_CODE_STREAM_CLOSE_TIMEOUT is set process-wide but never restored

The comment (lines 255-257) promises to "Save the original value so we can restore it in `finally` to avoid leaking state across concurrent jobs," and `savedStreamCloseTimeout` is declared (line 191) and assigned (line 260) for exactly that purpose. But no restore ever happens — the `finally` at lines 495-498 only stops the watcher, and `savedStreamCloseTimeout` is never read again. So the env var is mutated to `120000` process-wide and left set, and the tracking variable is effectively dead code.

Either drop the save-and-restore intent (and the unused variable) or actually restore the prior value in `finally`. Note the existing SDK path (`container-runner.ts` line 287-291) has the same set-without-restore, so if restoration is desired it should be added in both places.

packages/core/src/runner/runtime/container-runner.ts
```diff
@@ -83,18 +88,55 @@ export class ContainerRunner implements RuntimeInterface {
...
+
+    if (agent.mcp_servers) {
+      const [hostServers, containerServers] = partitionMcpServers(agent.mcp_servers);
+
+      if (Object.keys(hostServers).length > 0) {
+        // Start host-side MCP bridges
+        for (const [name, config] of Object.entries(hostServers)) {
+          const handle = await createMcpHostBridge(name, config);
+          hostBridges.push(handle);
+        }
```
edspencer, line 103: **LOW · bug** — Host bridges created before the try block leak if a later createMcpHostBridge throws

The loop that starts host bridges (lines 100-103) runs *before* the `try` whose `finally` closes them (lines 156-164). If `createMcpHostBridge` succeeds for the first server but throws for a later one (e.g. the host MCP process fails to connect / list tools), the already-created handles in `hostBridges` are never closed, leaking their subprocess + MCP client, and the error propagates out of `execute()` uncaught. Moving the bridge-startup loop inside the `try` (so the existing `finally` cleanup covers partial startup) would close this gap.

edspencer (commented):
This PR adds host-side MCP server support for Docker agents: a `host: true` schema flag, `partitionMcpServers` to split host vs container servers, `createMcpHostBridge` to spawn host MCP processes and expose their tools as injected server defs, and `mcpBridgeHost` routing so containerized agents reach bridges via `host.docker.internal`.

I read `cli-runtime.ts`, `container-runner.ts`, and `mcp-host-bridge.ts` in full, compared the new CLIRuntime bridge-startup block against the existing, correctly-cleaned-up `executeSDKRuntime` path, and checked the schema and tests. The partitioning logic, host-bridge tool delegation, and `host.docker.internal` routing look correct. However, the new CLIRuntime bridge/env code lacks the cleanup that its sibling SDK path has — details below.

edspencer:
Reviewed 6 changed files (14 hunks); ran the verify pass; 3 findings.

Issue #307, opened 2026-07-09
feat(core): reap streaming sessions on idle + re-trigger scheduled wakeups via the existing scheduler


edspencer:
## Summary

Persistent chat sessions (`openChatSession` / `SDKRuntime.openSession`) have **no lifecycle management** today — nothing closes the underlying native `claude` process, so a consumer driving every chat as a long-lived session (edspencer/paddock#111) would accumulate warm processes at **~300 MB RSS each**. This issue adds the missing lifecycle: **reap sessions the instant they go idle, unless they hold live background work**, and **re-trigger scheduled wakeups/crons through herdctl's existing scheduler** so cross-turn autonomy survives the reap.

This is a deliberately **minimal** design (revised down from an earlier idle-timer/backstop/concurrency-cap proposal — see the reasoning below): the whole thing is one decision at turn-end plus a small reuse of the scheduler we already have.

## Why so simple? (the economics)

Reaping is **cheap and lossless** (measured, SDK 0.3.205 / native 2.1.205):
- Resume recovers the full conversation (3/3 exact recall).
- Cold process spawn → `init` ≈ **0.5 s**; user-facing penalty on the next message ≈ **0.5–1.3 s** — hidden behind the LLM's own multi-second first token, on turns that run minutes-to-hours.
- **The Anthropic prompt cache is server-side and survives the reap** — a fresh resumed process got a **17,729-token cache _read_** within the 5-min TTL. Reaping loses no token cost, only the ~0.5 s respawn.

→ Keeping an idle session warm buys *only* ~1 s and costs ~300 MB. A time-based idle window (e.g. "reap after 5 min") would pin memory to save ~1 s on maybe one follow-up — a false economy in both memory and complexity. So: **no idle timer.** Reap the moment a session goes idle. The only sessions that stay warm are the ones actively doing work.

Bonus: with immediate reap there are **no warm idle sessions at all**, so steady-state memory self-bounds to *actively-working + continuous-class* sessions. No separate concurrency cap needed.

## Current state (`@herdctl/core@5.17.0`)

- `JobControl.openChatSession` hard-codes `new SDKRuntime()` and returns `runtime.openSession(...)` with **zero bookkeeping** (`packages/core/src/fleet-manager/job-control.ts` ~329, 388). It also does **not** thread `abortController` into the query (contrast `execute()`), so `close()`/`interrupt()` are the only levers.
[...]

PR #304, opened 2026-07-08
feat(core): bump claude-agent-sdk ^0.1.0 → ^0.3.x (fixes #303)

edspencer:
## Summary

Bumps `@anthropic-ai/claude-agent-sdk` in `packages/core` from `^0.1.0`
(resolved **0.1.77**) to `^0.3.0` (resolves **0.3.205**), closing #303.

The pinned `^0.1.0` range cannot cross a `0.x` minor, so herdctl was frozen on a
stale SDK line whose **bundled JS harness lacks the current agentic toolset**
(`ScheduleWakeup`, `ToolSearch`, `Cron*`, `Monitor`, …). `0.3.x` drops the
bundled harness and instead extracts/runs the **native Claude Code binary**,
which carries those tools — unblocking cross-turn autonomy for consumers driving
a persistent `openSession()` agent (e.g. an agent that schedules
`ScheduleWakeup` and is re-invoked when it fires).

## Changes

- `packages/core/package.json`: `@anthropic-ai/claude-agent-sdk` `^0.1.0` → `^0.3.0`; `pnpm-lock.yaml` updated.
- **`sdk-runtime.ts` adapter fixes** for 0.3.x type changes (the only two type breaks across the SDK surface used by `sdk-adapter` / `sdk-runtime` / `message-queue`):
  - `Query.interrupt()` now resolves to an optional interrupt-receipt object instead of `void`. `RuntimeSession.interrupt()` awaits and discards it to preserve its fire-and-forget `Promise<void>` contract.
  - The SDK's `tool()` handler return type now uses a literal-typed MCP `CallToolResult`. The injected-MCP adapter casts `InjectedMcpToolDef.handler` at that boundary, keeping the transport-agnostic tool definition SDK-free (no public-type change).
- Changeset added (`@herdctl/core`: minor).

## Verification

- `pnpm typecheck` — **11/11 pass**
- `pnpm build` — **7/7 pass**
- `pnpm lint` — **6/6 pass**
- Core tests — **3190 pass**, 29 skipped. The single failure (`directory.test.ts` "throws … when parent directory is not writable") is a **pre-existing, environment-only** failure that also fails on clean `main` when the test runner is `root` (root bypasses `chmod`); CI runs non-root, where it passes.
[...]

PR #308, opened 2026-07-09
feat(core): session-lifecycle reaper + wake registry (#307, part 1)

edspencer:
Implements the core of #307 — **reap idle streaming sessions the instant they go idle, and re-trigger the timer-class wakeups they leave behind through herdctl's own scheduler**. This is **part 1**: the complete, tested core mechanism + integration seams. Part 2 (the `FleetManager.openChatSession` opt-in + consumer delivery surface, coupled to edspencer/paddock#111) follows.

## Why
Today `openChatSession` opens an `SDKRuntime` streaming session (~300 MB RSS native `claude` process) with **zero lifecycle management** — nothing reaps it. A consumer driving every chat as a long-lived session (paddock#111) accumulates warm processes. Reaping is cheap and lossless (resume recovers full conversation; the Anthropic prompt cache is server-side and survives the reap), so the right policy is: **reap the moment a session goes idle, unless it holds live background work.** No idle timers, no backstops, no concurrency cap — memory self-bounds to *actively-working* sessions.

## What's in this PR (`packages/core/src/session/`)
- **`reaper-policy`** — `decideReap`: keep alive iff continuous-class `background_tasks` are running; otherwise reap. Timer-class `session_crons` never pin a session.
- **`wake-store`** — pure reconcile-by-id of SDK `session_crons` into herdctl-owned wake entries: absolute `nextRunAt` (via an injected cron resolver), one-shot vs recurring, **7-day recurring expiry**, and the **gap-4 recurring-ownership rule** (a herdctl-fired resumed turn doesn't re-arm a session-only cron, so recurring wakes are *not* dropped on absence — only on expiry or an explicit `CronDelete`).
- **`wake-registry`** — stateful coordinator: async-locked reconcile + **concurrency-limited (gap 2)**, deadlock-free `dispatchDue` that pre-persists post-fire state and **skips still-live sessions (gap 6)**.
- **`session-reaper`** — ties a live session to the policy; consumes `turn_end` / `background_tasks_changed` / `activity` signals; defers `close()` so it never re-enters the Stop hook; re-reaps a kept-alive session once its background work ends (and won't reap mid-turn).
- **`session-hooks`** — `buildLifecycleHooks` (`Stop`/`SubagentStop`) + `tapLifecycleStream` mapping SDK events → lifecycle signals.
- **`fleet-state-wake-persistence`** — durable `session_wakes` slice in fleet state.

### Integration seams
- `SDKRuntime.openSession` installs the lifecycle hooks + stream tap when a new `onLifecycleSignal` execute-option is set, and **threads an `AbortController`** (acceptance criterion).
- `Scheduler` gains an **`onTick`** option so wake firing reuses the existing loop (no second timer).

## Testing
- 47 new unit tests across the `session/` module + `Scheduler.onTick`.
- Core coverage **80%+** (all thresholds green); `src/session` ≈ 97%.
- `pnpm --filter @herdctl/core typecheck` + `build` green.
- Full suite green except the pre-existing **root-only** `directory.test.ts` (root bypasses the unwritable-dir check; unrelated).

## Not in this PR (part 2)
- `FleetManager` constructs the `WakeRegistry` (`resolveNextRun` via `cron.ts`, `fire` = `openChatSession({resume}) + inject`, `isSessionLive` from the reaper) and passes `onTick: () => registry.dispatchDue()` to the `Scheduler`.
- `ChatSessionOptions.manageLifecycle` opt-in in `JobControl.openChatSession` (wire `onLifecycleSignal` → `reaper.manage(session)`).
- Consumer surface for delivering the re-triggered turn onto the hub/attribution path (paddock#111 gap 3), and the targeted `CronDelete` watch (gap 4b).

Kept separate because it touches the large `FleetManager` and is coupled to the Paddock consumer surface — same rationale as splitting #304/#305.

Refs #307, edspencer/paddock#111.

## Summary by CodeRabbit

* **New Features**
  * Added support for long-lived session lifecycle handling, including automatic wake tracking and re-triggering.
  * Sessions can now report lifecycle events during streaming, enabling better background-work awareness.
  * Scheduler ticks can now run additional time-based work alongside existing schedule checks.

* **Bug Fixes**
  * Improved session teardown so open sessions close more reliably.
  * Preserved unrelated fleet state while updating session wake data.
  * Added safeguards so scheduler hook errors do not stop the main loop.

### DIFF
--- a/.changeset/session-reaper.md
+++ b/.changeset/session-reaper.md
@@ -1,0 +1,16 @@
+---
+"@herdctl/core": minor
+---
+
+Add streaming-session lifecycle management: a session reaper and a wake registry (part 1 of edspencer/herdctl#307).
+
+Long-lived chat sessions (`openChatSession` / `SDKRuntime.openSession`) now have the building blocks to be reaped the instant they go idle — rather than accumulating warm native `claude` processes at ~300 MB each — and to have their timer-class wakeups (`ScheduleWakeup` / `CronCreate`) re-triggered through herdctl's own scheduler after the reap.
+
+New, additive surface under `@herdctl/core`:
+
+- **`session/` module** — `decideReap` (the one-rule reap policy), `reconcileSessionWakes` + the wake-store helpers (reconcile SDK `session_crons` by id into durable wake entries with an absolute `nextRunAt`, one-shot vs recurring semantics, 7-day recurring expiry), `WakeRegistry` (async-locked, concurrency-limited, deadlock-free due-firing that skips still-live sessions), `SessionReaper` (drives reap vs keep-alive from turn-end / `background_tasks_changed` / activity signals), `buildLifecycleHooks` / `tapLifecycleStream`, and `FleetStateWakePersistence`.
+- **`SDKRuntime.openSession`** now installs `Stop`/`SubagentStop` lifecycle hooks and taps the message stream when a new `onLifecycleSignal` execute-option is provided, and threads an `AbortController` into the query for clean teardown.
+- **`Scheduler`** gains an `onTick` option so wake firing reuses the existing scheduler loop instead of a second timer.
+- Fleet state gains an optional `session_wakes` slice.
+
+No behavior changes to existing paths: sessions are only managed when a caller opts in via `onLifecycleSignal`, and the reaper/registry are not yet wired into `FleetManager.openChatSession` (that end-to-end wiring + consumer surface is part 2).
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -30,6 +30,8 @@ export * from "./hooks/index.js";
 export * from "./runner/index.js";
 // Scheduler exports (PRD 6)
 export * from "./scheduler/index.js";
+// Streaming-session lifecycle: reaper + wake re-trigger (#307)
+export * from "./session/index.js";
 // State exports (PRD 2)
 export * from "./state/index.js";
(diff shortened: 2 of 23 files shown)


edspencer:
CI green across all jobs (Lint, Typecheck, Unit Tests, Web UI, Build+Smoke). Marked draft pending the part-2 fleet wiring; ready for a look at the core mechanism whenever.

edspencer:
Addressed all three CodeRabbit findings in bf8822d (they're the same fire-and-forget/rejection theme):

- **session-reaper queue wedge (major):** `processSignal` now catches reconcile I/O failures and routes `onReap`/`onKeepAlive` through a guarded `notify()` so a throwing consumer callback can never skip `close()` (which would leak the very session we decided to reap). Applied your queue-separation suggestion too (`this.queue = result.then(undefined, () => undefined)`) as a backstop, so an unexpected throw can't poison the chain.
- **session-hooks `void sink(...)` (major + the sdk-runtime:205 dup):** both `tapLifecycleStream` sites and the Stop-hook callback now go through an `emit()` helper that observes the promise and logs rejections via `createLogger`, so a failing `onLifecycleSignal` can't become an unhandled rejection.

Added 5 tests: throwing `onReap` still closes; reap survives a failed reconcile; queue keeps processing after a failure; the hook resolves `{continue:true}` and the stream keeps flowing when the sink rejects. Full suite green (3204), coverage ~80%.

Issue #324, opened 2026-07-11
current_job/last_job never written; shutdown bulk-cancel reads a never-populated field


edspencer:
## Summary
`state.yaml`'s `current_job`/`last_job` agent fields are never written: `updateAgentState()` has zero production callers. Downstream, shutdown bulk-cancel collects running jobs from the never-populated `current_job` — making it effectively a no-op.

## Evidence
- Schema fields exist: `packages/core/src/state/schemas/fleet-state.ts:48-50`
- `updateAgentState()` (`packages/core/src/state/fleet-state.ts:194`) — no callers outside tests (repo-wide grep)
- Only *schedule* state is set to running (`packages/core/src/scheduler/scheduler.ts:571`)
- `cancelRunningJobs()` path reads each agent's `currentJobId` from the never-written field: `packages/core/src/fleet-manager/status-queries.ts:235`, used by `stop({cancelOnTimeout:true})` → `packages/core/src/fleet-manager/job-control.ts:675-698`

## Impact
- `stop()` with `cancelOnTimeout: true` cannot actually find/cancel running jobs.
- Anyone building dashboards or automation on `current_job`/`last_job` (as `architecture/job-system.md` instructs) gets nulls.

## Suggested fix
Write `current_job` at job start and `last_job` at completion (job-executor or ScheduleExecutor/JobControl), or drop the fields and re-source bulk-cancel from the in-memory AbortController registry (which now exists and is authoritative for this process: `job-control.ts:56-64`).

---
Found by the 2026-07-11 full-docs audit.

edspencer:
Partial fix up in #349.

**What's fixed:** the shutdown bulk-cancel no-op. `cancelRunningJobs()` was reading the `current_job` fleet-state field, which no production path ever writes, so it always cancelled nothing while agents kept running. In-flight jobs are now tracked in a single shared `AbortController` registry that **both** manual triggers and **scheduled** jobs register into — the latter is key, since scheduled jobs (the ones that actually stall shutdown) previously registered in no registry at all, so the "iterate the existing in-memory registry" approach alone would still have cancelled nothing. Shutdown now iterates the shared registry and genuinely aborts each live run; keyed by job id, so it's concurrency-safe under `max_concurrent > 1`.

[...]

Issue #328, opened 2026-07-11
job:created emitted after job completion; job:output never emitted for manual trigger()


edspencer:
## Summary
Mid-run job observability via events doesn't exist: `job:created` is emitted only after the job has finished (immediately before `job:completed`/`job:failed`), and `job:output` is emitted only for schedule-triggered jobs — never for manual `trigger()` runs.

## Evidence
- `job:created` post-completion: `packages/core/src/fleet-manager/job-control.ts:246-284`; `packages/core/src/fleet-manager/schedule-executor.ts:132-146` ("Emit job:created event now that we have the job info")
- `job:output` only from ScheduleExecutor: `schedule-executor.ts:114-129`; the `trigger()` path never emits it
- The only mid-run hook is the `onJobCreated` callback (`fleet-manager/types.ts`)
- `docs/src/content/docs/architecture/job-system.md` "Event Timeline" documents the opposite ordering (doc fix tracked separately, but the code behavior is the underlying surprise)

## Impact
Consumers subscribing to `job:created` to observe live jobs see nothing until the job ends — the same class of bug as Paddock #100 (in-flight chat invisible). `job:output` streaming recipes in the docs cannot work for manual triggers.

## Suggested fix
Emit `job:created` when the job record is created (both trigger paths), and emit `job:output` from the shared execution path rather than only ScheduleExecutor. Audit consumers for double-handling before changing ordering.

---
Found by the 2026-07-11 full-docs audit.

PR #349, opened 2026-07-11
fix(core): cancel in-flight jobs on shutdown via shared AbortController registry

edspencer:
## Summary

Fixes the shutdown bulk-cancel no-op reported in #324. `FleetManager.stop({ cancelOnTimeout: true })` (the path the CLI `stop` command uses) was a **guaranteed no-op**: on a scheduler shutdown timeout, `cancelRunningJobs()` collected running jobs from the `current_job` fleet-state field — but **no production code path ever writes `current_job`/`last_job`** (`updateAgentState()` has zero prod callers). So the set of "running jobs" was always empty, it logged `No running jobs to cancel`, and the agents kept running.

## Why the issue's suggested fix was incomplete

The issue suggested re-sourcing bulk-cancel from the in-memory AbortController registry in `JobControl` (`runningControllers`). But that registry **only tracked manual `trigger()` jobs** — `ScheduleExecutor.executeSchedule` called `JobExecutor.execute(...)` **without** an `abortController`, so scheduled jobs registered nowhere. Scheduled jobs are exactly the ones that stall shutdown, so a registry-only fix would *still* have cancelled nothing in the real scenario.

## Approach

Move the registry to a **single shared** one, hung off `FleetManagerContext` and implemented by `FleetManager`:

- New (optional) context accessors: `registerJob` / `unregisterJob` / `getJobController` / `getRunningJobIds`, backed by one `Map<jobId, AbortController>` on `FleetManager`. Optional to match the existing lifecycle-accessor convention so the lightweight mock contexts in the chat-manager unit tests need no changes; the real `FleetManager` always implements them.
- `JobControl.trigger` / `cancelJob` repointed at the shared registry (removing the dead `getAgentInfoFn` dependency).
- `ScheduleExecutor.executeSchedule` now creates an `AbortController`, registers it in `onJobCreated`, passes it into `executor.execute(...)`, unregisters in a `finally`, and skips the completed/failed emit on abort (mirroring `trigger()`'s cancellation handling).
- `cancelRunningJobs()` now iterates `getRunningJobIds()` and calls `cancelJob(jobId, { timeout })` per entry. Keyed by job id → concurrency-safe under `instances.max_concurrent > 1`.

## Deferred (out of scope — needs a design decision)

This PR fixes the **cancel** no-op but intentionally **leaves the `current_job`/`last_job` observability fields unwritten**. A single scalar `current_job` cannot represent concurrent jobs (`max_concurrent > 1`); whether to model it as an array vs. most-recent is an open schema question. Hence `Refs #324`, not `Fixes` — this only partially resolves the issue.

## Also noted (not fixed here)

Separate latent bug in the same shutdown path: `persistShutdownState` (`fleet-manager.ts`) writes `fleet.stoppedAt`, but `FleetMetadataSchema` only defines `started_at`, so Zod silently strips it. Flagging only — left for a separate change.

## Tests

- New regression test in `integration.test.ts`: a **scheduled** job genuinely in-flight at shutdown is now aborted via the registry (asserts the job's `AbortController` fired and the job is persisted `cancelled`). Verified it **fails without the fix** (scheduled-job registration disabled → `sawAbort` stays false).
- The existing "interrupts an in-flight job" `trigger()` test continues to pass through the moved registry.
- Full `@herdctl/core` suite: 3257 passed, 1 pre-existing failure only (`state/__tests__/directory.test.ts` "parent directory not writable" — a root-only chmod test, confirmed failing identically on clean `origin/main`). `discord`/`slack` suites green; core + dependents typecheck and build clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Shutdown cancellation now reliably stops in-progress manual and scheduled jobs when `cancelOnTimeout` is enabled.
  * Cancelled jobs now terminate active CLI processes and SDK requests, and are correctly recorded as cancelled.
  * Scheduled job completion and failure events are no longer emitted after a run has been cancelled.

### DIFF
--- a/.changeset/shutdown-bulk-cancel-registry.md
+++ b/.changeset/shutdown-bulk-cancel-registry.md
@@ -1,0 +1,19 @@
+---
+"@herdctl/core": patch
+---
+
+fix(core): shutdown bulk-cancel now actually cancels in-flight jobs
+
+`FleetManager.stop({ cancelOnTimeout: true })` was a guaranteed no-op: on a
+shutdown timeout it read jobs from a `current_job` fleet-state field that nothing
+ever wrote, so no jobs were ever cancelled. Manual and scheduled jobs that stalled
+shutdown kept running.
+
+In-flight jobs are now tracked in a single shared AbortController registry that
+both manual triggers (`JobControl.trigger`) and scheduled jobs
+(`ScheduleExecutor`) register into. Shutdown bulk-cancel iterates that registry and
+genuinely aborts each live run (killing the CLI subprocess / aborting the SDK
+query). The registry is keyed by job id, so it is concurrency-safe under
+`instances.max_concurrent > 1`. This is an internal, non-breaking change.
+
+Refs edspencer/herdctl#324.
--- a/packages/core/src/fleet-manager/__tests__/integration.test.ts
+++ b/packages/core/src/fleet-manager/__tests__/integration.test.ts
@@ -21,6 +21,7 @@ vi.mock("@anthropic-ai/claude-agent-sdk", () => ({
 
 // Import the mocked query function for test configuration
 import { query as mockQueryFn } from "@anthropic-ai/claude-agent-sdk";
+import { listJobs } from "../../state/index.js";
 import { AgentNotFoundError, InvalidStateError, ScheduleNotFoundError } from "../errors.js";
 import type {
   AgentStartedPayload,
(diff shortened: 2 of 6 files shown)


edspencer:
Rebased onto `main` to resolve conflicts with #348 (which merged first and touches the same `onJobCreated`/`onMessage` code in `job-control.ts` and `schedule-executor.ts`).

Both changes were combined rather than either winning:
- **onJobCreated** now does both — registers the AbortController in the shared registry (this PR, #324) **and** emits `job:created` at creation time with the in-hand `JobMetadata` (#348).
- **onMessage** uses #348's shared `buildJobOutputPayload` helper (the private `mapMessageTypeToOutputType`/`extractMessageContent` methods #324's branch referenced were removed by #348).
- Kept this PR's try/finally + `abortController.signal.aborted` early-return scaffolding around `execute()`.

Verified after rebase: `@herdctl/core` typecheck + build clean; full core suite 3297 passed / 29 skipped, only the known pre-existing root-only `directory.test.ts` chmod failure (unrelated). Both the #328 ordering test and the #324 in-flight-scheduled-job cancellation test pass together.

Issue #311, opened 2026-07-09
Session ScheduleWakeup resolves ~24h late when host TZ ≠ UTC (local-cron vs UTC-resolver mismatch)


edspencer:
## Summary

A session-only `ScheduleWakeup` (a *relative* "wake me in N seconds/minutes") gets its next-run time resolved **~24h in the future** when the host's local timezone is behind UTC. The SDK harness serializes the one-shot wake as a minute/hour **cron expression in the harness's LOCAL time**, but the session reaper's wake-store cron resolver evaluates that cron in **UTC**. When local ≠ UTC and the local wall-clock minute/hour has already passed in UTC today, `nextRunAt` rolls to tomorrow — so a "1 minute from now" wake effectively never fires (until the next day).

Found while wiring the reaper into a consumer (edspencer/paddock#111). The Paddock-side capture/persistence/dispatch all worked; the wake just resolved to the wrong absolute time.

## Reproduction

Host timezone `America/New_York` (UTC-4). Real wall-clock: `2026-07-09 19:08 EDT` = `23:08 UTC`.

A keeper in a managed streaming session calls `ScheduleWakeup` for ~60s out. The persisted `session_wakes` entry (herdctl state):

```yaml
session_wakes:
  d137639f:
    id: d137639f
    agent: keeper-autotest
    sessionId: 82262a46-...
    schedule: "10 19 * * *"                 # 19:10 — the LOCAL target minute/hour
    recurring: false
    prompt: "…"
    nextRunAt: 2026-07-10T19:10:00.000Z      # resolved as 19:10 UTC → already past today → tomorrow
    createdAt: 2026-07-09T23:08:46.380Z      # 23:08 UTC == 19:08 local
```

`schedule: "10 19 * * *"` is 19:10 **local** (60s from 19:08 local). The resolver treated `19:10` as UTC; since it's currently `23:08 UTC` (past `19:10 UTC`), `nextRunAt` became **tomorrow** `19:10 UTC`. The wake sits idle for ~20h instead of firing in a minute.

Forcing `nextRunAt` to an overdue value and restarting made the reaper fire it immediately and correctly (resume-and-inject → the woken turn ran), confirming the *only* defect is the timezone used to resolve the cron.

## Root cause (likely)

Two independent timezone assumptions that don't agree:
1. The SDK/native harness expresses a relative one-shot `ScheduleWakeup` as a `SessionCronSummary.schedule` minute/hour cron in **local** time.
[...]

PR #379, opened 2026-07-18
feat(core): scheduler host-execution seam + runtime schedule mutation (#375, #376)

### LINKED ISSUE #375: Scheduler: host-execution seam (setScheduleTriggerHandler) + headless fallback + TZ hardening
**D1** · unblocks paddock#265 (scheduled chats).

Today a scheduled fire runs a fresh **headless** job via `ScheduleExecutor`/`JobExecutor`, off any host hub (`handleScheduleTrigger`, fleet-manager.ts:1240). An embedding host (Paddock) needs to own execution so the run streams into its UI. Add a seam mirroring the existing `setSessionWakeHandler` (fleet-manager.ts:952):

```ts
private async handleScheduleTrigger(info: TriggerInfo): Promise<void> {
  if (this.scheduleTriggerHandler) { await this.scheduleTriggerHandler(info); return; }
  await this.scheduleExecutor.executeSchedule(info); // unchanged fallback
}
```

Plus `setScheduleTriggerHandler(handler)`. No Scheduler-core change — `interval`/`cron` timing is unchanged; we only intercept execution. When a host handler runs the turn via its own resume path, the `--resume`/`isSidechain` hiding problem is bypassed entirely.

Also harden the relative-wake **timezone** seam (#311) and confirm 'trigger now' flows through the handler path.

**Blocks:** paddock#265.

---
Part of the Paddock **Events / Schedules / Config** initiative. Additive + backward-compatible — no behaviour change for existing headless fleets. Anchors captured against herdctl main; re-verify before implementing.

### LINKED ISSUE #376: Scheduler: runtime add/remove of a single schedule + state prune + mutation gate
**D2** · unblocks paddock#265/#266 (adding schedules from a UI at runtime).

herdctl can enable/disable a schedule at runtime but has **no API to add or remove a single schedule** on a registered agent — only whole-agent `addAgent(replace:true)`, which **leaves stale state**: removed schedules' persisted `last_run_at`/status linger, a `disabled` status keeps winning after replace (state is read, not config, scheduler.ts:374), and orphaned in-memory scheduler-map entries survive.

Add, mirroring `enable/disableSchedule` + `AgentManagement.commit` (arming already reads live `agent.schedules` each tick, so no Scheduler-core change):
- `FleetManager.setAgentSchedule(agent, name, schedule)` — validate vs `ScheduleSchema`, set the key, re-push via `scheduler.setAgents(...)`.
- `FleetManager.removeAgentSchedule(agent, name)` — delete key, re-push, **prune** persisted state (new `deleteScheduleState` in scheduler/schedule-state.ts) + clear in-memory `runningSchedules`/`warnedSchedules`.
- **Per-deployment gate** controlling whether programmatic schedule mutation is allowed at all (none exists today; schedules run with zero gating — this is the opt-in 'security' control).

**Blocks:** paddock#265, paddock#266.

---
Part of the Paddock **Events / Schedules / Config** initiative. Additive + backward-compatible — no behaviour change for existing headless fleets. Anchors captured against herdctl main; re-verify before implementing.

edspencer:
Makes herdctl's scheduler host-friendly so an embedding host (Paddock) can **own execution** and **mutate schedules at runtime**. Implements **#375 (D1)** and **#376 (D2)** together — they touch the same files (fleet-manager, schedule-management, scheduler, schedule-state), so splitting would only conflict with itself.

All changes are **additive and backward-compatible** — existing headless fleets behave exactly as before; every new capability is opt-in.

## D1 (#375) — host-execution seam
- **`FleetManager.setScheduleTriggerHandler(handler)`**, mirroring `setSessionWakeHandler`. When set, `handleScheduleTrigger` routes a fired schedule to the host handler (`await handler(info); return;`) and skips the built-in headless `ScheduleExecutor`; when unset, schedules run headless as before. No Scheduler-core timing change — cron/interval scheduling is untouched, only the *execution* of a due trigger is intercepted. Every scheduler-fired trigger (including a forced immediate fire) funnels through the seam.
- **Relative-wake timezone hardening (#311):** new `resolveSystemTimeZone()` centralizes the host-timezone lookup used by `calculateNextCronTrigger`/`calculatePreviousCronTrigger`, falling back to `UTC` when an ICU-less runtime reports no timezone (which could otherwise revive the 24h-idle bug).

## D2 (#376) — runtime schedule mutation
- **`FleetManager.setAgentSchedule(agent, name, schedule)`** — validates against `ScheduleSchema`, sets the key on the stored agent, and re-pushes via `scheduler.setAgents(...)`. Arming reads live `agent.schedules` each tick, so the schedule is immediately eligible to fire.
- **`FleetManager.removeAgentSchedule(agent, name)`** — deletes the key, re-pushes, **prunes persisted state** via a new `deleteScheduleState` helper in `scheduler/schedule-state.ts`, and clears the in-memory `runningSchedules`/`warnedSchedules` entries via a new `Scheduler.clearScheduleTracking`. Without the prune, a removed schedule's `last_run_at`/`disabled` status would linger and a re-added name would silently inherit it (the scheduler reads state, not config, to decide disabled/next-run).
- **Per-deployment mutation gate:** new `allowScheduleMutation` FleetManager option (default `false`). Both mutation methods throw the new `ScheduleMutationDisabledError` when it's not enabled. Enable/disable of existing schedules stays ungated. This is the opt-in safety control — there was none before.

Mirrors the existing `enableSchedule`/`disableSchedule` + `AgentManagement.commit` patterns throughout.

## Tests
- Seam calls the host handler when set, and falls back to headless when not (plus clearing the handler restores headless).
- `setAgentSchedule` arms a new schedule on an already-running agent (the scheduler fires it), validates input, and rejects unknown agents.
- `removeAgentSchedule` prunes persisted state so a re-added name starts clean (no stale `last_run_at`/`disabled`).
- The mutation gate blocks both methods when disabled, leaves enable/disable ungated, and allows mutation when opted in.
- Unit coverage for `deleteScheduleState` and `resolveSystemTimeZone` (incl. the ICU-less UTC fallback).

Full `@herdctl/core` suite green except one pre-existing, environment-specific failure unrelated to this change (`directory.test.ts` "parent directory not writable" — the assertion depends on `chmod` denying writes, which doesn't hold when tests run as root; fails identically on clean `main`). Typecheck, full monorepo build, and docs build all pass. Docs (library-reference + error-handling) and a changeset (`@herdctl/core` minor) are included.

Closes #375
Closes #376

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in host execution hook for scheduler-fired schedules, including forced/immediate triggers, with headless fallback when unset.
  * Added runtime APIs to set/replace or remove a single agent schedule (programmatic add/remove after registration).
  * Added an opt-in `allowScheduleMutation` gate (default off) for runtime schedule edits.
* **Bug Fixes**
  * Prevented resurrection of deleted/stale schedule state and improved safety for concurrent schedule mutations.
  * Hardened timezone handling with deterministic UTC fallback in timezone resolution.
* **Documentation**
  * Documented the new option, APIs, execution handler, and the `ScheduleMutationDisabledError` (including error code).

### DIFF
--- a/packages/core/src/fleet-manager/schedule-management.ts
+++ b/packages/core/src/fleet-manager/schedule-management.ts
@@ -2,15 +2,26 @@
  * Schedule Management Module
  *
  * Centralizes all schedule management logic for FleetManager.
- * Provides methods to query, enable, and disable schedules.
+ * Provides methods to query, enable, disable, add, and remove schedules.
  *
  * @module schedule-management
  */
 
+import {
+  type ResolvedAgent,
+  type ResolvedConfig,
+  type Schedule,
+  ScheduleSchema,
+} from "../config/index.js";
 import type { FleetManagerContext } from "./context.js";
-import { AgentNotFoundError, ScheduleNotFoundError } from "./errors.js";
+import {
+  AgentNotFoundError,
+  ConfigurationError,
+  ScheduleMutationDisabledError,
+  ScheduleNotFoundError,
+} from "./errors.js";
 import { buildScheduleInfoList, type FleetStateSnapshot } from "./status-queries.js";
-import type { ScheduleInfo } from "./types.js";
+import type { ConfigChange, ScheduleInfo } from "./types.js";
 
 // =============================================================================
 // ScheduleManagement Class
@@ -19,13 +30,20 @@ import type { ScheduleInfo } from "./types.js";
 /**
  * ScheduleManagement provides all schedule management operations for the FleetManager.
  *
- * This class encapsulates the logic for querying, enabling, and disabling schedules
- * using the FleetManagerContext pattern.
+ * This class encapsulates the logic for querying, enabling, disabling, adding, and
+ * removing schedules using the FleetManagerContext pattern.
+ *
+ * Runtime add/remove of a single schedule ({@link setAgentSchedule} /
+ * {@link removeAgentSchedule}) is gated behind `allowScheduleMutation` — supplied
+ * as {@link isMutationAllowed} — so a deployment must opt in before schedules can
+ * be mutated programmatically (edspencer/herdctl#376).
  */
 export class ScheduleManagement {
   constructor(
     private ctx: FleetManagerContext,
     private readFleetStateSnapshotFn: () => Promise<FleetStateSnapshot>,
+    private setConfig: (config: ResolvedConfig) => void,
+    private isMutationAllowed: () => boolean,
   ) {}
 
   /**
@@ -204,5 +222,186 @@ export class ScheduleManagement {
 
     // Return the updated schedule info
     return this.getSchedule(agent.qualifiedName, scheduleName);
+  }
+
+  /**
+   * Add or replace a single schedule on an already-registered agent at runtime.
+   *
+   * Unlike {@link enableSchedule}/{@link disableSchedule} (which flip persisted
+   * state), this changes the agent's *configuration*: the schedule is validated
+   * against {@link ScheduleSchema}, set on the stored agent's `schedules` map, and
+   * the updated agent list is re-pushed to the scheduler. Arming reads live
+   * `agent.schedules` each tick, so the schedule is immediately eligible to fire
+   * (or, when re-registering an existing name, picks up the new timing) without a
+   * `reload()` or whole-agent `addAgent(replace)` round trip. A `config:reloaded`
+   * event describing the change is emitted.
+   *
+   * Gated behind `allowScheduleMutation` (see {@link ScheduleMutationDisabledError}).
+   *
+   * @param agentName - Qualified or local name of an existing agent
+   * @param scheduleName - The schedule key to set
+   * @param schedule - The schedule definition (validated against ScheduleSchema)
+   * @returns The resulting schedule info
+   * @throws {ScheduleMutationDisabledError} If schedule mutation is not enabled
+   * @throws {AgentNotFoundError} If the agent doesn't exist
+   * @throws {ConfigurationError} If the schedule fails validation
+   */
+  async setAgentSchedule(
+    agentName: string,
+    scheduleName: string,
+    schedule: Schedule | Record<string, unknown>,
+  ): Promise<ScheduleInfo> {
+    this.assertMutationAllowed("setAgentSchedule");
+
+    const logger = this.ctx.getLogger();
+    const { config, agent } = this.requireAgent(agentName);
+
+    // Validate against the schema so runtime input gets the same defaults
+    // (e.g. enabled/resume_session) and coercions a file-loaded schedule would.
+    let validated: Schedule;
+    try {
+      validated = ScheduleSchema.parse(schedule);
+    } catch (error) {
+      throw new ConfigurationError(
+        `Invalid schedule configuration for "${agentName}/${scheduleName}": ` +
+          `${error instanceof Error ? error.message : String(error)}`,
+        { cause: error instanceof Error ? error : undefined },
+      );
+    }
+
+    const isReplace = Boolean(agent.schedules && scheduleName in agent.schedules);
+    const updatedAgent: ResolvedAgent = {
+      ...agent,
+      schedules: { ...(agent.schedules ?? {}), [scheduleName]: validated },
+    };
+
+    this.commitAgent(config, agent, updatedAgent, {
+      type: isReplace ? "modified" : "added",
+      category: "schedule",
+      name: `${agent.qualifiedName}/${scheduleName}`,
+      details: validated.type,
+    });
+
+    logger.info(
+      `${isReplace ? "Updated" : "Added"} schedule ${agent.qualifiedName}/${scheduleName} programmatically`,
+    );
+
+    return this.getSchedule(agent.qualifiedName, scheduleName);
+  }
+
+  /**
+   * Remove a single schedule from an agent at runtime.
+   *
+   * Deletes the schedule key from the stored agent's `schedules` map, re-pushes
+   * the updated agent list to the scheduler, and — critically — prunes the
+   * schedule's persisted state via `deleteScheduleState` and clears the in-memory
+   * scheduler bookkeeping (`runningSchedules`/`warnedSchedules`) via
+   * `Scheduler.clearScheduleTracking`. Without the prune, a later re-add of the
+   * same name would inherit the removed schedule's stale `last_run_at`/`disabled`
+   * status, because the scheduler decides disabled/next-run from state, not config
+   * (edspencer/herdctl#376). A `config:reloaded` event is emitted.
+   *
+   * Gated behind `allowScheduleMutation` (see {@link ScheduleMutationDisabledError}).
+   *
+   * @param agentName - Qualified or local name of an existing agent
+   * @param scheduleName - The schedule key to remove
+   * @returns `true` if a schedule was removed, `false` if the agent had no such schedule
+   * @throws {ScheduleMutationDisabledError} If schedule mutation is not enabled
+   * @throws {AgentNotFoundError} If the agent doesn't exist
+   */
+  async removeAgentSchedule(agentName: string, scheduleName: string): Promise<boolean> {
+    this.assertMutationAllowed("removeAgentSchedule");
+
+    const logger = this.ctx.getLogger();
+    const stateDir = this.ctx.getStateDir();
+    const { config, agent } = this.requireAgent(agentName);
+
+    if (!agent.schedules || !(scheduleName in agent.schedules)) {
+      logger.debug(
+        `removeAgentSchedule: no schedule "${scheduleName}" on agent "${agent.qualifiedName}"`,
+      );
+      return false;
+    }
+
+    const { [scheduleName]: _removed, ...remainingSchedules } = agent.schedules;
+    const updatedAgent: ResolvedAgent = { ...agent, schedules: remainingSchedules };
+
+    this.commitAgent(config, agent, updatedAgent, {
+      type: "removed",
+      category: "schedule",
+      name: `${agent.qualifiedName}/${scheduleName}`,
+    });
+
+    // Prune persisted state so a re-added name doesn't inherit stale
+    // last_run_at / disabled status (edspencer/herdctl#376).
+    const { deleteScheduleState } = await import("../scheduler/schedule-state.js");
+    await deleteScheduleState(stateDir, agent.qualifiedName, scheduleName, {
+      logger: { warn: logger.warn },
+    });
+
+    // Clear in-memory scheduler bookkeeping keyed by this schedule.
+    this.ctx.getScheduler()?.clearScheduleTracking(agent.qualifiedName, scheduleName);
+
+    logger.info(`Removed schedule ${agent.qualifiedName}/${scheduleName} programmatically`);
+    return true;
+  }
+
+  // ===========================================================================
+  // Private Helpers
+  // ===========================================================================
+
+  /**
+   * Throw {@link ScheduleMutationDisabledError} unless the deployment has opted
+   * into programmatic schedule mutation.
+   */
+  private assertMutationAllowed(operation: string): void {
+    if (!this.isMutationAllowed()) {
+      throw new ScheduleMutationDisabledError(operation);
+    }
+  }
+
+  /**
+   * Resolve an agent by qualified name (preferred) or local name from the loaded
+   * config, throwing {@link AgentNotFoundError} if absent.
+   */
+  private requireAgent(agentName: string): { config: ResolvedConfig; agent: ResolvedAgent } {
+    const config = this.ctx.getConfig();
+    const agents = config?.agents ?? [];
+    const agent =
+      agents.find((a) => a.qualifiedName === agentName) ?? agents.find((a) => a.name === agentName);
+
+    if (!config || !agent) {
+      throw new AgentNotFoundError(agentName, {
+        availableAgents: agents.map((a) => a.qualifiedName),
+      });
+    }
+
+    return { config, agent };
+  }
+
+  /**
+   * Replace one agent in the stored config with an updated copy, push the new
+   * agent list to the scheduler, and emit a `config:reloaded` event. Mirrors
+   * `AgentManagement.commit`, the single seam for programmatic config mutation.
+   */
+  private commitAgent(
+    config: ResolvedConfig,
+    previous: ResolvedAgent,
+    updated: ResolvedAgent,
+    change: ConfigChange,
+  ): void {
+    const newAgents = config.agents.map((a) => (a === previous ? updated : a));
+    const newConfig: ResolvedConfig = { ...config, agents: newAgents };
+    this.setConfig(newConfig);
+
+    this.ctx.getScheduler()?.setAgents(newAgents);
+
+    this.ctx.emit("config:reloaded", {
+      agentCount: newAgents.length,
+      agentNames: newAgents.map((a) => a.qualifiedName),
+      configPath: config.configPath,
+      changes: [change],
+      timestamp: new Date().toISOString(),
+    });
   }
 }
--- a/packages/core/src/fleet-manager/schedule-management.ts
+++ b/packages/core/src/fleet-manager/schedule-management.ts
@@ -254,6 +254,7 @@ export class ScheduleManagement {
     this.assertMutationAllowed("setAgentSchedule");
 
     const logger = this.ctx.getLogger();
+    const stateDir = this.ctx.getStateDir();
     const { config, agent } = this.requireAgent(agentName);
 
     // Validate against the schema so runtime input gets the same defaults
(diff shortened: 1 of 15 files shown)


packages/core/src/fleet-manager/schedule-management.ts
```diff
@@ -205,4 +223,185 @@ export class ScheduleManagement {
...
+      category: "schedule",
+      name: `${agent.qualifiedName}/${scheduleName}`,
+      details: validated.type,
+    });
+
+    logger.info(
+      `${isReplace ? "Updated" : "Added"} schedule ${agent.qualifiedName}/${scheduleName} programmatically`,
+    );
+
+    return this.getSchedule(agent.qualifiedName, scheduleName);
```
edspencer, line 301: **LOW · correctness** — setAgentSchedule replacing a disabled schedule leaves it disabled despite "immediately eligible to fire" contract

`setAgentSchedule` mutates only config (`commitAgent` → `setAgents`) and never touches persisted schedule state. Whether a schedule fires is decided by persisted state: the scheduler skips any schedule whose state `status === "disabled"` (scheduler.ts:395). So if a schedule was previously disabled (via `disableSchedule`) and a host then calls `setAgentSchedule` on that same name to replace/update it, the new definition inherits the lingering `disabled` state and will **not** fire — even though the method's doc contract states the schedule "is immediately eligible to fire (or picks up new timing when replacing an existing name)". `getSchedule` returned from this method will also report `status: "disabled"`.

The `removeAgentSchedule` path deliberately prunes persisted state to avoid exactly this stale-state inheritance; the replace path in `setAgentSchedule` has no equivalent, so the two mutation methods are asymmetric. This is an edge case (only when replacing a name that is currently disabled), hence low severity, but it contradicts the documented behavior and could silently leave a "set" schedule dormant. Consider either clearing/resetting persisted state for the replaced name, or documenting that `setAgentSchedule` does not re-enable a disabled schedule (caller must `enableSchedule`).

edspencer (commented):
This PR adds two backward-compatible, opt-in capabilities to the scheduler: a host-execution seam (`setScheduleTriggerHandler`) that lets an embedding host own execution of fired schedules, and runtime single-schedule mutation (`setAgentSchedule`/`removeAgentSchedule`) gated behind a new `allowScheduleMutation` option, plus a UTC-fallback timezone helper (`resolveSystemTimeZone`). The changes are cleanly additive and well-tested.

I verified: the seam is read live in `handleScheduleTrigger` and invoked inside the scheduler's `executeJob` try/catch, so a throwing host handler is recorded as `last_error` rather than becoming an unhandled rejection; the mutation methods have no `await` between reading and re-committing config, so the in-memory read-modify-write is not interleaved (single-threaded); `deleteScheduleState` follows the existing `updateScheduleState` read-write pattern; the scheduler skips disabled schedules based on persisted state (scheduler.ts:395) and arms live from `agent.schedules` each tick. One behavioral gap noted below around `setAgentSchedule` replacing a disabled schedule.

edspencer:
Diffed against the prior revision: the substantive change is the running-gated tombstone lift in scheduler.ts `setAgents` plus `armScheduleState` no longer deleting the tombstone, with matching test/doc updates. Verified the tombstone invariant end-to-end (set only when wasRunning; guaranteed lift via the in-flight run's executeJob finally which runs after — and thus does not un-suppress — the trailing writes; setAgents lift is a no-op backstop). Confirmed the running-set retention still skips a concurrent same-name re-add as "already running" (maxInFlight stays 1). Export-surface narrowing from the prior revision is intact.

Reviewed 10 changed files (23 hunks); 0 findings.

packages/core/src/scheduler/schedule-state.ts
```diff
@@ -136,42 +189,181 @@ export async function updateScheduleState(
...
+  scheduleName: string,
+  options: ScheduleStateOptions = {},
+): Promise<boolean> {
+  const stateFilePath = getStateFilePath(stateDir);
+
+  return withStateLock(stateFilePath, async () => {
+    // Tombstone first — even if there is nothing on disk to prune yet, an
+    // in-flight run may still be about to write. The tombstone makes that write
+    // a no-op until the schedule is re-armed.
+    tombstonedSchedules.add(tombstoneKey(stateFilePath, agentName, scheduleName));
```
edspencer: **MEDIUM · correctness** — Tombstone is only lifted by setAgentSchedule; a removed schedule re-introduced via reload()/addAgent stays permanently suppressed

`deleteScheduleState` adds the schedule to the process-global `tombstonedSchedules` set, and `updateScheduleState` becomes a silent no-op for any tombstoned key (lines 195-198). The tombstone is lifted **only** in `armScheduleState`, which is called **only** from `setAgentSchedule`. Any other path that re-introduces the same `agent/schedule` name never clears it:

- `removeAgentSchedule` → `commitAgent` mutates only in-memory config (schedule-management.ts:429-433) and does **not** rewrite the on-disk config file. So after a runtime remove, the schedule still exists on disk.
- A subsequent `reload()` calls `loadConfiguration()` (re-reads the unchanged on-disk config) and `scheduler.setAgents(...)` (config-reload.ts:101-105), re-arming the schedule — but the tombstone is still set.
- `addAgent({ replace: true })` with an agent carrying the same schedule name has the same effect.

Once re-armed while tombstoned, every `updateScheduleState` for that schedule is dropped. Arming reads unlocked `getScheduleState`, which now returns default state (`last_run_at: null`) because the prune deleted the entry; `calculateNextTrigger(null, …)` returns `now` (interval.ts:142-144), so the schedule is always "due." The trailing idle/next-run write after each fire is suppressed, so `last_run_at` never persists — an interval schedule then re-fires on essentially every scheduler tick (runaway execution), and no schedule state is ever recorded. A cron schedule similarly re-fires throughout its trigger window and never records `last_run_at`/`next_run_at`.

The happy path (remove → re-add via `setAgentSchedule`) is fine and tested, but the remove-then-`reload()` sequence is realistic for a long-running embedding host (removal is in-memory only, so a resync/reload brings the schedule back). Secondary: because tombstones are never cleared except by re-arm, the `tombstonedSchedules` set also grows unbounded across repeated removals.

Suggest tying the tombstone's lifetime to the in-flight run's completion (e.g. clear it once no run is in flight for the key, or when `setAgents` re-introduces the name) rather than solely to `setAgentSchedule`, so `reload()`/`addAgent` re-armed schedules resume persisting state. Reported as a correctness issue for the maintainer to weigh; the fix spans multiple sites so I've described it rather than presume an exact edit.

edspencer (commented):
Revised PR. It now (a) resolves my earlier finding by adding `armScheduleState` to normalize a lingering persisted `disabled` status to `idle` on `setAgentSchedule`, and (b) adds concurrency coordination to schedule-state: a per-state-file promise-chain lock (`withStateLock`) that serializes read-modify-writes to prevent lost sibling updates, plus a process-global tombstone set that suppresses an in-flight run's trailing write from resurrecting a just-deleted schedule. `clearScheduleTracking` was narrowed to only clear warn-once bookkeeping (retaining the in-flight running-set entry), with `isScheduleRunning` added for logging.

I traced the tombstone lifecycle end-to-end: it is only ever added in `deleteScheduleState` and only ever removed in `armScheduleState` (called exclusively from `setAgentSchedule`). I verified `removeAgentSchedule`/`commitAgent` mutate only in-memory config (no on-disk yaml write), that `reload()` re-reads the unchanged config from disk and re-arms via `setAgents`, that unlocked `getScheduleState` returns default state (last_run_at null) after a prune, and that `calculateNextTrigger(null, …)` fires immediately (interval.ts:142). That combination surfaces the one correctness issue below. The lock itself (`then(fn, fn)` with a non-throwing tail) looks correct, and the documented "one trailing idle write after same-name re-add" residual is acknowledged and bounded.

edspencer:
Addressed the second-round MEDIUM (tombstone leak → runaway firing after reload) in `d67582c`.

**Root cause:** the process-global tombstone was only cleared by `armScheduleState` (i.e. only via `setAgentSchedule`). So `removeAgentSchedule` (in-memory) followed by `reload()` / `addAgent({replace:true})` re-armed the same schedule **from disk while still tombstoned** → every `updateScheduleState` no-op'd → `last_run_at` never persisted → schedule read as perpetually due → runaway per-tick firing. The set also grew unbounded.

**Fix — bounded, universally-cleared tombstone:**
- **`setAgents` is the universal un-tombstone hook.** Every re-arm path (`reload()`, `addAgent({replace})`, `setAgentSchedule`) funnels through `scheduler.setAgents`, so it now lifts the tombstone for every schedule **present** in the new agent list. A schedule brought back from disk resumes normal state persistence instead of staying tombstoned. (A schedule *absent* from the list keeps its tombstone — it may still have an in-flight run to guard.)
- **Lifetime bounded to the run.** `executeJob`'s `finally` clears the tombstone when the in-flight run completes, and `removeAgentSchedule` only tombstones when a run is actually in flight — when nothing is running there's no writer to guard against, so no tombstone is set and it can't leak.
- **No await-gap race.** Tombstoning moved out of `deleteScheduleState` into a synchronous `Scheduler.tombstoneSchedule()` that `removeAgentSchedule` calls with no `await` in between, so a concurrent re-arm's `setAgents` can't clear it before it's set.

**Ambiguity resolved:** `removeAgentSchedule` is **in-memory-only**, consistent with `addAgent`/`removeAgent` — it does not rewrite `herdctl.yaml`. A later `reload()` legitimately brings the schedule back, and now does so cleanly. Documented on the method and in the changeset.

**Test added:** remove → `reload()` re-introduces the schedule from disk → asserts it resumes state persistence (`last_run_at` recorded), the tombstone is cleared, and it does **not** runaway-fire (fires ≤ 1 over ~10 ticks). Plus updated tombstone/arm units for the explicit-tombstone API.

Local: typecheck clean, biome clean, full monorepo + docs build pass, full `@herdctl/core` suite green except the one pre-existing root-only `directory.test.ts` env failure (unrelated; `chmod` can't deny writes when tests run as root). Watching CI now.

packages/core/src/scheduler/index.ts
```diff
@@ -33,11 +34,16 @@ export {
 } from "./schedule-runner.js";
 // Schedule state management
 export {
+  armScheduleState,
+  clearScheduleTombstone,
```
edspencer: **LOW · maintainability** — Internal tombstone mutators are exported on the public @herdctl/core surface (semver footgun)

`scheduler/index.ts` now exports `setScheduleTombstone` and `clearScheduleTombstone` (plus `isScheduleTombstoned`), and the package root `src/index.ts` does `export * from "./scheduler/index.js"`, so these become part of the public, semver-versioned `@herdctl/core` API (the changeset is a `minor` bump, and CLAUDE.md marks this a LIVE PROJECT that follows semver).

These are internal concurrency-coordination primitives that are only correct when paired with the exact `removeAgentSchedule` sequence (synchronous set-before-prune, lifted by `setAgents`/`executeJob` finally). Exposed publicly they are a footgun: a consumer calling `setScheduleTombstone(stateDir, agent, name)` directly would make every subsequent `updateScheduleState` for that schedule a silent no-op until a re-arm lifts it — `last_run_at` never persists, the schedule reads as perpetually due, and it runaway-fires every tick (precisely the failure mode this PR works to prevent). Publishing them also commits the project to maintaining them under semver.

Since they appear to be exported mainly for the unit tests, consider keeping the mutating helpers (`setScheduleTombstone`/`clearScheduleTombstone`) module-internal — tests in the same package can import them directly from `./schedule-state.js` without re-exporting through `scheduler/index.ts` and the package root. `armScheduleState`/`deleteScheduleState` are legitimately part of the runtime-mutation feature and reasonable to export.

edspencer (commented):
Third revision. It resolves both of my earlier findings: (1) `setAgentSchedule` now normalizes a lingering persisted `disabled` status to `idle` via `armScheduleState`, and (2) the previous tombstone-leak / runaway-firing risk is closed — `deleteScheduleState` no longer tombstones unconditionally; instead `removeAgentSchedule` sets the tombstone synchronously only when a run is actually in flight (`isScheduleRunning`), and the tombstone is lifted on every re-arm path (`setAgents` clears it for every schedule present in an updated agent list, covering `reload()`/`addAgent`/`setAgentSchedule`) and by the run's own `executeJob` finally. A new test exercises the remove→reload() path and asserts no per-tick runaway.

I traced the full lifecycle: `triggerSchedule` adds to `runningSchedules` synchronously before its first await (so `isScheduleRunning` has no TOCTOU gap against `checkAllSchedules`, which calls `triggerSchedule` with no intervening await after `shouldTrigger`); the per-file `withStateLock` serializes the tombstone-check + prune + trailing writes so the final state converges to "pruned" regardless of interleaving; and the tombstone Set is now bounded (cleared on re-arm and on run completion; not set at all when nothing is in flight). The remaining "one trailing idle write after same-name re-add" residual is explicitly documented and is bounded/self-correcting. The design looks sound. One minor API-surface note below.

edspencer:
Addressed the LOW (internal API leaking onto the public surface) in `4440e3f` — done before merge so it doesn't ship in the release.

`scheduler/index.ts` no longer re-exports the runtime-mutation state helpers (`setScheduleTombstone`, `clearScheduleTombstone`, `isScheduleTombstoned`, and — same category — `armScheduleState`, `deleteScheduleState`). Since the package root does `export *`, those would otherwise have landed on the public `@herdctl/core` semver surface. This restores the pre-PR public schedule-state surface (`getScheduleState` / `getAgentScheduleStates` / `updateScheduleState` + types).

Internal callers (`scheduler`, `schedule-management`) and the schedule-state tests already import these directly from `./schedule-state.js`; the one test that reached `isScheduleTombstoned` via the index now imports it from the source module. External consumers use the `FleetManager` methods (`setAgentSchedule` / `removeAgentSchedule`).

Verified against the built `dist`: no public `.d.ts` re-exports any of the five (the only remaining mention is a `{@link}` JSDoc reference). Typecheck + biome clean; full `@herdctl/core` suite green except the one pre-existing root-only `directory.test.ts` env failure (unrelated). Should be clean to merge/release now.

edspencer (commented):
Fourth revision. This iteration addresses my prior (low) finding about API-surface hygiene: the internal runtime-mutation helpers (`armScheduleState`, `deleteScheduleState`, `setScheduleTombstone`, `clearScheduleTombstone`, `isScheduleTombstoned`) are no longer re-exported from `scheduler/index.ts`, so they stay off the public `@herdctl/core` semver surface (which does `export *` from that module). The test that needs `isScheduleTombstoned` now imports it directly from `./schedule-state.js`.

I grepped every reference to these symbols: internal callers reach them directly (`scheduler.ts` imports from `./schedule-state.js`; `schedule-management.ts` uses `await import("../scheduler/schedule-state.js")`) and none go through the index or package root, so removing the re-exports breaks no import — and since these functions are all new in this PR, no external consumer depended on them. The core concurrency machinery (per-file `withStateLock`, the running-gated synchronous tombstone, `setAgents` un-tombstoning present schedules, and the `executeJob` finally clear) is unchanged from the previous revision I already traced and validated. No outstanding issues.

edspencer:
Addressed the Major (tombstone cleared too early → old-run write contaminates the re-add) in `8665ae0`. This is the generation/epoch coordination the `removeAgentSchedule` path had been circling — it unifies both directions rather than contradicting the earlier fix.

**Unified rule:** clear the tombstone on re-arm (fixes runaway), **but not while the removed generation's run is still in flight** (prevents contamination).

- `scheduler.setAgents` now lifts the tombstone for a re-armed schedule **only when no run for that key is in flight** (gated on `isScheduleRunning`). While the removed gen's run is still executing, the tombstone is preserved so its trailing `next_run_at`/`last_error` write stays suppressed.
- `executeJob`'s `finally` lifts the tombstone when the in-flight run completes (after its trailing writes) — the coordinated hand-off point, after which the re-armed schedule persists cleanly.
- `armScheduleState` no longer unconditionally clears the tombstone; clearing is owned by `setAgents`/`executeJob`. (Removal prunes state, so a re-add finds no entry there and is a no-op anyway.)

**Test:** remove + re-add while a gen-0 run is gated now asserts the tombstone **remains** across the re-add, that gen-0's failure (`last_error`) does **not** persist onto the re-added schedule, that the tombstone is lifted once gen-0 completes, and no double execution (`maxInFlight === 1`). Updated the `armScheduleState` unit to assert it does not lift the tombstone.

Kept targeted — no ballooning, so no need to split the D1 seam out. Local: typecheck + biome clean, full `@herdctl/core` suite green except the one pre-existing root-only `directory.test.ts` env failure. Watching CI.

edspencer (commented):
Fifth revision. Since the last iteration it tightens the tombstone/execution-generation coordination: `setAgents` now lifts a re-armed schedule's tombstone only when no run for that key is in flight (previously unconditional), and `armScheduleState` no longer lifts the tombstone at all. This closes the previously-deferred residual where a still-in-flight removed generation's trailing `last_error`/`next_run_at` write could contaminate a freshly re-added same-name schedule.

I read the on-disk `setAgents`, `executeJob` finally, `removeAgentSchedule`, and the arm/tombstone helpers, and traced the lifecycle: a tombstone is only ever set when `removeAgentSchedule` observes `wasRunning === true` (synchronously, no await gap), so a concrete in-flight run always exists whose `executeJob` `finally` unconditionally clears that exact key — a guaranteed backstop. Therefore the conditional lift in `setAgents` cannot strand a tombstone (no reintroduction of the runaway per-tick firing I flagged earlier), while suppression correctly holds for the removed generation's trailing writes until that run finishes. Both of my earlier findings remain resolved and the new tests exercise both the contamination and reload-runaway paths. No outstanding issues.

PR #391, opened 2026-07-20
fix(core): re-establish injected MCP servers on session wakes (#390)

### LINKED ISSUE #390: Session wake/resume path drops injectedMcpServers → in-process MCP tools vanish for the whole autonomous stretch
## Summary

Under session drive-mode (SDK runtime), **in-process injected MCP servers passed via `injectedMcpServers` are dropped on every wake-driven turn** (`ScheduleWakeup` / `/loop` / `CronCreate` re-fires). A keeper that goes autonomous therefore loses its injected MCP tools for the entire autonomous stretch — the tools are still listed in `--allowedTools` but are **unbacked** (no server behind the pattern), so they vanish from the model's catalog. They only return when a caller re-drives the session with `injectedMcpServers` again (e.g. an interactive/human turn). Observed episodes last **several hours**; worst seen ~9–10 h continuous, across multiple projects. After a host/process restart the loss becomes effectively **permanent** for affected chats, because the durable wake set re-fires without injection.

## Root cause

`injectedMcpServers` is a per-call option. It reaches the runtime only when a caller passes it into `openChatSession` → `openSession`:

`fleet-manager/job-control.ts`
```ts
const session = runtime.openSession({
  agent: effectiveAgent,
  prompt: options?.prompt ?? "",
  resume: sessionId,
  injectedMcpServers: options?.injectedMcpServers,   // undefined on the wake path
  ...
});
```

But the wake-registry fire path re-opens the session with no such option, and the options type has no slot for it:

`session/session-lifecycle-manager.ts`
```ts
/** Options a fired wake passes to `openChatSession`. */
export interface SessionWakeChatOptions {
  resume: string;
  prompt: string;
  manageLifecycle: true;
}

private async fire(entry: SessionWakeEntry): Promise<void> {
  const session = await this.openChatSession(entry.agent, {
    resume: entry.sessionId,
    prompt: entry.prompt,
    manageLifecycle: true,          // ← no injectedMcpServers
  });
  if (this.sessionWakeHandler) { await this.sessionWakeHandler(session, entry); return; }
  ...
}
```

A consumer that registers a `sessionWakeHandler` **cannot** repair this: `fire()` calls `openChatSession` (spawning the MCP-less subprocess) **before** invoking the handler, so the handler only receives an already-opened, injection-less session to drain.

`injectedMcpServers` is never persisted on the session, the `SessionWakeEntry`, or the `ResolvedAgent`, so the wake path has nothing to reconstruct it from. Durable wakes (`FleetStateWakePersistence`) re-fire after a restart the same way → the tools stay gone until a fresh injected turn.

### Reproduction (shape)
1. Open a managed chat session with a non-empty `injectedMcpServers` (an in-process `createSdkMcpServer`) — tools present.
2. Have the turn leave a `ScheduleWakeup`/`CronCreate` and go idle so the reaper closes it.
3. Let the wake fire. The resumed turn's tool catalog no longer contains the injected server's tools; only tools whose servers are re-supplied by the caller survive. All subsequent wake turns are identical.

## Secondary (cosmetic, but a clear fingerprint): `allowedTools` accumulates duplicates

On each SDK turn the injected-server tool patterns are appended to a **shared reference** of the agent's persistent `allowed_tools` array, so `--allowedTools` grows unbounded (`mcp__<name>__*` repeated 2×, 3×, 4×…):

- `runner/sdk-adapter.ts` assigns `result.allowedTools = agent.allowed_tools` (by reference, no copy).
- `runner/runtime/sdk-runtime.ts` then `sdkOptions.allowedTools.push(\`mcp__${name}__*\`)` in place, every turn.

Fix: copy the array in the adapter (`result.allowedTools = [...agent.allowed_tools]`) and de-dupe before pushing in the runtime.

## Proposed fix (primary)

Let the wake path re-establish injected servers. Add `injectedMcpServers` — or, cleaner, a `resolveInjectedMcpServers(entry) => Record<string, InjectedMcpServerDef>` factory — to `SessionLifecycleManagerOptions` / `SessionWakeChatOptions`, and pass it through `fire()` into `openChatSession`. A consumer registers the factory alongside its wake handler (keyed by agent + sessionId), reusing its normal injected-server builders. Alternatively, cache the `injectedMcpServers` supplied at a managed session's first open, keyed by sessionId, and re-apply on every subsequent wake fire of that session (no consumer change needed). This closes both the live-flap and the post-restart-durable-wake cases.

## Impact

Any consumer that injects in-process MCP servers into session-mode agents and relies on cross-turn autonomy (`ScheduleWakeup`/`/loop`) loses those tools for the whole autonomous run. Declared schedule/hook triggers that re-drive through the consumer's own turn-start path are unaffected — only herdctl-owned session wakes exhibit the loss.

Traced against `@herdctl/core` at `chore: version packages (#373)`.

edspencer:
Fixes #390.

## The bug

Under session drive-mode (SDK runtime), in-process **`injectedMcpServers`** (in-memory `createSdkMcpServer` servers, e.g. a consumer's self-management/file-sender MCP) are dropped on every **wake-driven** turn — `ScheduleWakeup` / `/loop` / `CronCreate` re-fires.

The human/first turn passes them through `openChatSession` → `openSession` and the tools work. But when the session goes idle the reaper closes it, and the wake registry re-fires it **without** `injectedMcpServers`: the resumed `claude` subprocess still lists the injected `mcp__…__*` patterns in `--allowedTools` but has **no server behind them**, so the tools vanish from the model's catalog for the entire autonomous stretch (observed multi-hour episodes). Because the wake set is durable, after a process restart the surviving wakes re-fire the same MCP-less way → the loss is effectively permanent for that chat until a fresh injected (human/consumer) turn.

A registered `sessionWakeHandler` can't repair it: `fire()` opens the session **before** invoking the handler, so the handler only receives an already-MCP-less session. And `injectedMcpServers` was never persisted anywhere the wake path could reconstruct it from.

## The fix (shape chosen: resolver factory)

I went with the **resolver/factory** shape rather than cache-at-first-open, because it's the more general and testable one and keeps injection *policy* (depth/gating, the actual server builders) in the consumer while herdctl owns the *wiring*. Cache-at-first-open would have needed `injectedMcpServers` plumbed into `manage()` (the first open happens outside the lifecycle manager), a more invasive change for no extra generality.

- Add an optional `resolveInjectedMcpServers?: (entry: SessionWakeEntry) => Record<string, InjectedMcpServerDef> | undefined` to `SessionLifecycleManagerOptions`, plus a `setResolveInjectedMcpServers` setter on both `SessionLifecycleManager` and `FleetManager` (mirrors the existing `setSessionWakeHandler` wiring).
- `fire()` consults the resolver (keyed off the wake's agent + sessionId) and threads the result into `openChatSession` via a new optional `injectedMcpServers` field on `SessionWakeChatOptions` — which flows straight through to `openSession`.

**Backward-compatible:** with no resolver registered, wakes fire with no injection exactly as before. A resolver that throws is logged and degrades to no-injection rather than wedging the wake.

## Secondary fix — `allowedTools` wildcard accumulation

The injected `mcp__…__*` patterns were accumulating on `--allowedTools` (repeated 2×/3×/4× per turn) because:

- `sdk-adapter.ts` assigned `result.allowedTools = agent.allowed_tools` **by reference** — a live pointer to the long-lived `ResolvedAgent`'s array. Now copied: `[...agent.allowed_tools]`.
- `sdk-runtime.ts` **pushed** `mcp__${name}__*` onto that array in place each turn. Now de-dupes before pushing.

## Tests

- **`session-lifecycle-manager.test.ts`** — wake fire forwards resolved `injectedMcpServers` into `openChatSession`; fires without injection (no crash) when no resolver is registered; `setResolveInjectedMcpServers` swaps at runtime; a throwing resolver degrades to no-injection without wedging the wake.
- **`sdk-runtime-allowed-tools.test.ts`** (new) — drives the real `execute()` path (mocked SDK `query()`): the injected wildcard is added exactly once, does not accumulate across repeated turns on the same agent object, does not mutate the persistent agent's array, and is de-duped when the agent already lists it.
- **`sdk-adapter.test.ts`** — `toSDKOptions` returns a fresh `allowedTools` array (not aliased to `agent.allowed_tools`).

Full `@herdctl/core` suite is green (one pre-existing, environment-only failure in `state/directory.test.ts` that asserts a non-writable-parent mkdir rejects — it fails identically on unmodified `main` when the suite runs as root, since root bypasses the permission; unrelated to this change and passes in CI). Added a changeset (`@herdctl/core` patch).

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored in-process MCP server availability when sessions resume after wake-triggered turns.
  * Preserved MCP tools during autonomous periods and across restarts.
  * Prevented duplicate MCP tool entries from accumulating across turns.
  * Wake operations now continue safely without MCP injection if resolution fails.

* **New Features**
  * Added support for dynamically resolving MCP servers when waking sessions.

### DIFF
--- a/.changeset/wake-injected-mcp-390.md
+++ b/.changeset/wake-injected-mcp-390.md
@@ -1,0 +1,27 @@
+---
+"@herdctl/core": patch
+---
+
+Re-establish injected MCP servers on session wake-fired turns (#390).
+
+Under session drive-mode, in-process `injectedMcpServers` were dropped on every
+wake-driven turn (`ScheduleWakeup` / `/loop` / `CronCreate` re-fires): the reaper
+closes the idle session and the wake registry re-opens it without the in-process
+servers, so the resumed subprocess still lists the injected `mcp__…__*` patterns
+in `--allowedTools` but has no server behind them — the tools vanish from the
+model's catalog for the whole autonomous stretch (durably so across restarts).
+
+- **@herdctl/core:** add an optional `resolveInjectedMcpServers(entry)` factory to
+  `SessionLifecycleManagerOptions` (and a `setResolveInjectedMcpServers` setter on
+  both `SessionLifecycleManager` and `FleetManager`). When registered, the wake
+  `fire()` path consults it and threads the resolved servers into the resumed
+  session via a new optional `injectedMcpServers` field on
+  `SessionWakeChatOptions`. Injection *policy* stays in the consumer; herdctl owns
+  the *wiring*. Fully backward-compatible — with no resolver registered, wakes fire
+  with no injection exactly as before, and a throwing resolver is logged and
+  degrades to no-injection rather than wedging the wake.
+
+- **@herdctl/core (secondary):** stop `allowedTools` from accumulating duplicate
+  `mcp__…__*` wildcards across turns. `toSDKOptions` now copies the agent's
+  `allowed_tools` array instead of aliasing it, and the SDK runtime de-dupes
+  injected tool patterns before appending them.
--- a/packages/core/src/fleet-manager/fleet-manager.ts
+++ b/packages/core/src/fleet-manager/fleet-manager.ts
@@ -27,7 +27,11 @@ import {
 import type { RuntimeSession, SlashCommand } from "../runner/index.js";
 import { getCliSessionFile, getDockerSessionFile } from "../runner/runtime/cli-session-path.js";
 import { Scheduler, type TriggerInfo } from "../scheduler/index.js";
-import { SessionLifecycleManager, type SessionWakeHandler } from "../session/index.js";
+import {
+  type ResolveInjectedMcpServers,
+  SessionLifecycleManager,
+  type SessionWakeHandler,
+} from "../session/index.js";
 import {
   type ChatMessage,
   type DiscoveredSession,
@@ -108,6 +112,7 @@ export class FleetManager extends EventEmitter implements FleetManagerContext {
   private scheduler: Scheduler | null = null;
   private sessionLifecycle: SessionLifecycleManager | null = null;
   private sessionWakeHandler?: SessionWakeHandler;
+  private resolveInjectedMcpServers?: ResolveInjectedMcpServers;
   private scheduleTriggerHandler?: ScheduleTriggerHandler;
 
   // Timing info
@@ -993,6 +998,7 @@ export class FleetManager extends EventEmitter implements FleetManagerContext {
       stateDir: this.stateDir,
       openChatSession: (agent, opts) => this.openChatSession(agent, opts),
       sessionWakeHandler: this.sessionWakeHandler,
+      resolveInjectedMcpServers: this.resolveInjectedMcpServers,
     });
   }
 
@@ -1006,6 +1012,26 @@ export class FleetManager extends EventEmitter implements FleetManagerContext {
   setSessionWakeHandler(handler: SessionWakeHandler | undefined): void {
     this.sessionWakeHandler = handler;
     this.sessionLifecycle?.setSessionWakeHandler(handler);
+  }
+
+  /**
+   * Register (or clear) the factory that re-supplies a session wake's in-process
+   * injected MCP servers.
+   *
+   * herdctl does not persist the per-call `injectedMcpServers` a consumer passes
+   * into {@link openChatSession}, so when the reaper closes an idle session and
+   * the wake registry re-fires it, the resumed `claude` subprocess loses those
+   * in-process servers — the injected `mcp__…__*` tools stay in `--allowedTools`
+   * but have no server behind them and vanish from the model's catalog for the
+   * whole autonomous stretch (edspencer/herdctl#390). A consumer (e.g. Paddock)
+   * registers this factory alongside {@link setSessionWakeHandler}, keyed off the
+   * wake's agent + sessionId, to rebuild the same servers on each fire. Without
+   * it, wakes fire with no injection (the pre-existing behavior). Safe to call
+   * before or after initialize().
+   */
+  setResolveInjectedMcpServers(resolve: ResolveInjectedMcpServers | undefined): void {
+    this.resolveInjectedMcpServers = resolve;
+    this.sessionLifecycle?.setResolveInjectedMcpServers(resolve);
   }
 
   /**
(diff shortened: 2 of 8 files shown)


.changeset/wake-injected-mcp-390.md
```diff
@@ -0,0 +1,27 @@
+---
+"@herdctl/core": patch
```
edspencer, line 2: **LOW · maintainability** — New public API suggests a minor bump, not patch

This changeset declares a `patch`, but the PR adds new backward-compatible **public API surface** to `@herdctl/core`: a new public method `FleetManager.setResolveInjectedMcpServers(...)` (fleet-manager.ts:1032), the exported `ResolveInjectedMcpServers` type, and a new `injectedMcpServers` field / `resolveInjectedMcpServers` option on the session lifecycle types. Under SemVer, adding functionality in a backward-compatible manner is a **minor** bump, not a patch. The repo emphasizes this explicitly (root CLAUDE.md "LIVE PROJECT - Follow Semver" plus the dedicated `changesets.md` rule).

The underlying #390 bug fix alone would be a patch, but the release will also ship a new opt-in consumer API, so consumers depending on that surface would expect it under a minor. Consider bumping to `minor` unless the maintainer deliberately intends to treat the new API as an internal-only patch.

edspencer (commented):
Fixes the loss of in-process injected MCP servers on wake-driven turns by adding a consumer-registered `resolveInjectedMcpServers` factory that `fire()` consults and threads through `openChatSession` → `openSession`, plus a secondary fix for `allowedTools` wildcard accumulation (copy the agent array in `toSDKOptions`; de-dupe before pushing in the SDK runtime).

I read `session-lifecycle-manager.ts`, `sdk-runtime.ts`, `sdk-adapter.ts`, and `fleet-manager.ts` in full around the diff, and traced the `injectedMcpServers` data path end-to-end: `fire()` (SLM 184-206) → fleet-manager `openChatSession` wiring (999) → job-control `openSession` (336/413) → sdk-runtime `toSDKOptions` (277-314). Confirmed the array is now a copy so the persistent `ResolvedAgent.allowed_tools` is never mutated, the de-dupe prevents unbounded wildcard growth turn-over-turn, the resolver throw is caught so it can't wedge a wake, and the setter propagates to a live `sessionLifecycle` or is captured at construction. Types line up (`SessionWakeChatOptions`/`ChatSessionOptions` both carry `injectedMcpServers`). Tests exercise the real paths. The logic is sound; only a semver/changeset classification nit below.

edspencer:
Reviewed 8 changed files (17 hunks); ran the verify pass; 1 finding.

PR #396, opened 2026-07-21
feat(core): persist per-run per-model token accounting + SDK cost on job record

### LINKED ISSUE #378: [P3] Persist per-run per-model token accounting + SDK cost on the job record
**X1** · P3; unblocks paddock#271 (cost) and run-history cost columns.

herdctl already receives the SDK's authoritative `total_cost_usd` + `usage` on the result message (message-processor.ts:427) and aggregates CLI-runtime tokens (cli-runtime.ts:571) — then **discards all of it** before the terminal `updateJob` (job-executor.ts:744). The job record (JobMetadataSchema, state/schemas/job-metadata.ts:47) carries no token/cost fields.

Persist per-run token accounting at run-end, **per model** (each model's four token classes — a chat can span an Opus main + Haiku subagents), + `num_turns` + SDK `total_cost_usd` when present. Keep accounting as accurate as possible. Pricing stays in the consuming app (Paddock) — do **not** bake a $/token table into herdctl's persisted state (it drifts; Max/CLI has no real per-token spend).

**Blocks:** paddock#271 (X2).

---
Part of the Paddock **Events / Schedules / Config** initiative. Additive + backward-compatible — no behaviour change for existing headless fleets. Anchors captured against herdctl main; re-verify before implementing.

edspencer:
Closes #378. Unblocks paddock#271.

## What

herdctl already receives the SDK's authoritative `total_cost_usd` + `usage` on the terminal result message (and the CLI runtime aggregates its own token counts), but **discarded all of it** before the final `updateJob`. This persists that accounting on the job record.

The job record (`JobMetadataSchema`) gains an additive, backward-compatible `usage` field capturing, at run-end:

- **`usage.per_model`** — per-model token accounting keyed by model id. Each entry carries that model's **four token classes** (`input_tokens`, `output_tokens`, `cache_creation_input_tokens`, `cache_read_input_tokens`). A single run can span multiple models (e.g. an Opus main agent delegating to Haiku subagents), so counts are broken down per model rather than summed.
- **`usage.num_turns`** — total agentic turns reported by the runtime.
- **`usage.total_cost_usd`** — the SDK's own authoritative run cost, when reported. Absent for CLI / Max-plan runs that have no real per-token spend.

**Pricing stays out of herdctl** — only raw token counts and the SDK's own cost figure are stored (no `$/token` table, which drifts). The consuming app (Paddock) applies pricing.

## How

- **`state/schemas/job-metadata.ts`** — new `ModelTokenUsageSchema` + `RunUsageSchema`; `usage: RunUsageSchema.nullable().optional()` on `JobMetadataSchema`; factory initializes it to `null`.
- **`runner/message-processor.ts`** — new `extractRunUsage(message)` maps the SDK result's camelCase `modelUsage` (with a top-level snake_case `usage` fallback keyed by model id, else `"unknown"`) into the persisted shape, coercing malformed counts to `0`.
- **`runner/runtime/cli-runtime.ts`** — the synthesized CLI result now aggregates **per-model** usage (incl. cache classes) into a `modelUsage` map, so both runtimes persist a uniform breakdown.
- **`runner/job-executor.ts`** — captures accounting from the terminal result message and writes it in the final `updateJob` (left untouched — stays `null` — when a run produces no result).

## Backward compatibility

Additive only. Existing job records without the field still parse; headless fleets see no behaviour change. Verified by a schema test that parses a legacy record with no `usage` key.

## Testing

- **Schema** — `RunUsageSchema` / `ModelTokenUsageSchema` validation (multi-model, optional cost/turns, empty map), legacy-record backward-compat, populated round-trip.
- **`extractRunUsage`** — SDK `modelUsage`, top-level `usage` fallback, `unknown` keying, malformed→0 coercion, CLI-synthetic shape, empty/none → undefined.
- **CLI runtime** — per-model aggregation across Opus + Haiku turns (incl. cache classes), `unknown` bucket for model-less turns.
- **End-to-end (real run path)** — `job-executor` test asserts the **persisted job record on disk** carries `usage.per_model` / `num_turns` / `total_cost_usd`, and stays `null` with no result message.

`pnpm build` + `pnpm typecheck` pass. Full core suite green except one pre-existing failure (`directory.test.ts` "parent not writable" — fails on clean main too; a root-permission-bypass in this sandbox, passes on CI's non-root runner).

### DIFF
--- a/.changeset/persist-token-accounting.md
+++ b/.changeset/persist-token-accounting.md
@@ -1,0 +1,13 @@
+---
+"@herdctl/core": minor
+---
+
+Persist per-run, per-model token accounting and SDK cost on the job record
+
+herdctl already receives the SDK's authoritative `total_cost_usd` + `usage` on the terminal result message and aggregates CLI-runtime tokens, but discarded all of it before the final `updateJob`. The job record now carries a `usage` field capturing this accounting at run-end:
+
+- **Per model** (`usage.per_model`, keyed by model id) — each model's four token classes (`input_tokens`, `output_tokens`, `cache_creation_input_tokens`, `cache_read_input_tokens`). A single run can span multiple models (e.g. an Opus main agent delegating to Haiku subagents), so counts are broken down per model rather than summed.
+- **`usage.num_turns`** — total agentic turns reported by the runtime.
+- **`usage.total_cost_usd`** — the SDK's own authoritative run cost, when reported. Absent for CLI / Max-plan runs that have no real per-token spend.
+
+New `extractRunUsage` reader maps the SDK result's camelCase `modelUsage` to the persisted snake-case shape; the CLI runtime now aggregates per-model usage (incl. cache classes) into its synthesized result so both runtimes persist a uniform breakdown. Pricing stays out of herdctl — only raw token counts and the SDK's own cost figure are stored; the consuming app applies a price table. Additive and backward-compatible: existing job records without the field still parse, and headless fleets see no behaviour change.
--- a/packages/core/src/runner/__tests__/job-executor.test.ts
+++ b/packages/core/src/runner/__tests__/job-executor.test.ts
@@ -176,6 +176,93 @@ describe("JobExecutor", () => {
       expect(job?.duration_seconds).toBeGreaterThanOrEqual(0);
     });
 
+    it("persists per-model token accounting from the result message", async () => {
+      const messages: SDKMessage[] = [
+        { type: "system", content: "Start" },
+        { type: "assistant", content: "Working" },
+        {
+          type: "result",
+          subtype: "success",
+          result: "Task completed!",
+          num_turns: 4,
+          total_cost_usd: 0.2571,
+          usage: { input_tokens: 10, output_tokens: 20 },
+          // SDK per-model breakdown: an Opus main agent + a Haiku subagent.
+          modelUsage: {
+            "claude-opus-4-8": {
+              inputTokens: 800,
+              outputTokens: 400,
+              cacheCreationInputTokens: 100,
+              cacheReadInputTokens: 6000,
+              webSearchRequests: 0,
+              costUSD: 0.25,
+              contextWindow: 200000,
+            },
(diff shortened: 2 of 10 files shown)


edspencer (commented):
Persists per-run, per-model token accounting (four token classes, num_turns, SDK `total_cost_usd`) on the job record via a new additive, backward-compatible `usage` field. The design correctly stores raw counts only and leaves pricing to the consumer. I reviewed the schema, `extractRunUsage`, the CLI-runtime aggregation, and the job-executor capture point, and traced the exports — no blocking issues found.

edspencer:
Checked:
- `state/schemas/job-metadata.ts`: `ModelTokenUsageSchema`/`RunUsageSchema` and the `usage: RunUsageSchema.nullable().optional()` field; factory inits `null`. Nonnegative constraints and optionality line up with the reader's output.
- `message-processor.ts::extractRunUsage`: guards non-object/non-result; prefers camelCase `modelUsage`, falls back to snake-case top-level `usage` keyed by `model` or `"unknown"`; coerces malformed counts to 0 via `toTokenCount`; returns `undefined` when nothing worth persisting. Logic and the `Object.keys(perModel).length === 0` fallback gate are consistent with the tests.
- `cli-runtime.ts`: per-model camelCase aggregation with dedup by message id and `stop_reason === null` filtering (pre-existing), summed into a synthetic result — same shape `extractRunUsage` consumes. Token summing across turns matches the prior behavior for input/output tokens.
- `job-executor.ts`: captures usage only from `type === "result"` messages ("later wins"), and spreads `usage` into the final `updateJob` only when present, leaving it `null` otherwise.
- Verified `ModelTokenUsage`/`RunUsage` are re-exported through `state/index.ts` (`export * from "./schemas/index.js"`), so the `message-processor` import resolves.

Tests cover multi-model, fallback, `unknown` bucketing, malformed→0 coercion, backward-compat legacy parse, and the on-disk job-executor round trip. No injection attempts in the PR text.

Reviewed 10 changed files (22 hunks); 0 findings.

PR #397, opened 2026-07-21
fix(core): prune stale adhoc session metadata entries (#168)

### LINKED ISSUE #168: Bug: Unbounded growth of adhoc.json session metadata — no pruning of stale entries
## Summary

The `adhoc.json` session metadata file (stored at `.herdctl/session-metadata/adhoc.json`) grows unboundedly because there is no pruning, TTL, or cleanup mechanism for stale metadata entries. On production (devbox.valfenda.net), there are 110 adhoc session metadata entries that are never cleaned up.

**This is separate from #164** (attribution race condition for new sessions) and **distinct from #167** (cross-agent metadata contamination). Issue #167 addresses metadata written to the *wrong agent's file*; this issue addresses the fact that the `adhoc.json` file itself grows indefinitely with entries for sessions that may no longer exist on disk.

### How adhoc metadata entries are created

When `SessionDiscoveryService.getAllSessions()` scans `~/.claude/projects/`, directories that don't match any known fleet agent use `metadataKey = "adhoc"` (line 552 of `session-discovery.ts`). Every session in these unattributed directories gets an autoName and preview extracted from its JSONL file and cached into `.herdctl/session-metadata/adhoc.json` via `batchSetAutoNames("adhoc", ...)` and `batchSetPreviews("adhoc", ...)`.

The critical amplification path: `getAllSessionGroups()` in `web-chat-manager.ts` calls `getAllSessions()` **without a limit parameter**, so Phase 2's top-N filtering is skipped and every single non-sidechain session in every non-temp directory is enriched and has metadata written.

### Why entries are never removed

The `SessionMetadataStore` class has **no cleanup, pruning, TTL, or expiry logic whatsoever**. The only "delete" operation is `removeCustomName()`, which removes a user-set custom name but doesn't address the autoName/preview cache entries. There is:

- No TTL or expiry check on metadata entries
- No periodic pruning job
- No max-entries limit
- No reconciliation against existing JSONL files
- No staleness detection

### Growth pattern

Every unique Claude Code CLI session file discovered in any unattributed directory gets an entry added to `adhoc.json`. These entries persist even after:
- The underlying JSONL session file is deleted by Claude Code
- The working directory no longer exists
- The session hasn't been accessed in months

On a development machine with active Claude Code usage, new sessions are created frequently, and old ones are never cleaned up, leading to monotonic growth.

### Key files

| File | Line | Role |
|------|------|------|
| `packages/core/src/state/session-metadata.ts` | Entire file | `SessionMetadataStore` — no cleanup logic exists |
| `packages/core/src/state/session-discovery.ts` | 552 | `metadataKey = agentName ?? "adhoc"` — adhoc key assignment |
| `packages/core/src/state/session-discovery.ts` | 660-666 | Batch writes metadata for all enriched sessions |
| `packages/web/src/server/chat/web-chat-manager.ts` | 211 | `getAllSessions(agentList)` called without limit — enriches ALL sessions |

## Impact

**Severity: Medium** — Not a crash or data loss bug, but causes degradation over time:

1. **Growing file size**: `adhoc.json` grows with every newly discovered unattributed session. At 110 entries with autoName + autoNameMtime + preview + previewMtime per entry, the file is still manageable (~50-100KB), but on machines with heavy CLI usage it will grow to thousands of entries over months.

2. **Increasing I/O per request**: Every call to `getAllSessions()` (triggered by `/api/chat/all` and `/api/chat/recent`) reads the entire `adhoc.json` into memory, parses it, and potentially writes it back with new entries. As the file grows, this read/parse/write cycle gets slower.

3. **Memory overhead**: The in-memory cache in `SessionMetadataStore` holds the full parsed `adhoc.json` object, which grows proportionally.

4. **Wasted JSONL parsing**: The autoName/preview cache entries reference sessions whose JSONL files may no longer exist. When `resolveAutoName()` checks the cache and finds a valid entry, it skips the JSONL parse — but the entry itself is dead weight that was created by parsing a file that no longer matters.

## Proposed Fix

Add a `prune(agentName, validSessionIds: Set<string>)` method to `SessionMetadataStore` that removes entries whose sessionId is not in the provided set. Call it at the end of `getAllSessions()` for each directory group, passing the set of sessionIds that were actually found during the scan.

```typescript
// In SessionMetadataStore
async prune(agentName: string, validSessionIds: Set<string>): Promise<number> {
  const metadata = await this.loadMetadata(agentName);
  if (!metadata) return 0;

  let pruned = 0;
  for (const sessionId of Object.keys(metadata.sessions)) {
    if (!validSessionIds.has(sessionId)) {
      delete metadata.sessions[sessionId];
      pruned++;
    }
  }

  if (pruned > 0) {
    await this.saveMetadata(agentName, metadata);
  }
  return pruned;
}
```

Additionally, consider:
- A configurable max-entries limit (e.g., keep most recent 500 entries by mtime)
- Only running the prune on every Nth call or at startup to avoid prune-on-every-request overhead

## Relationship to other issues

- **#164** (attribution race condition): Separate issue. #164 is about new sessions not appearing; this is about old sessions never being purged.
- **#167** (cross-agent metadata contamination): Related but distinct. #167 addresses metadata being written to the wrong agent's file. Even after #167 is fixed, adhoc.json would still grow unboundedly for genuinely unattributed sessions.

edspencer:
## What

Adds `SessionMetadataStore.prune(agentName, validSessionIds)` and wires it into `SessionDiscoveryService.getAllSessions()` so stale session-metadata entries are reconciled away against the sessions that still exist on disk.

## Why

`SessionMetadataStore` cached `autoName` / `preview` / `isSidechain` / `usage` entries per session under a metadata key (`"adhoc"` for unattributed sessions) but had **no** cleanup, TTL, or reconciliation. Every unique Claude Code session ever discovered in an unattributed directory got an entry in `adhoc.json`, and entries persisted forever — even after the underlying JSONL transcript was deleted. In production this had grown to 110 dead entries, and the file grows monotonically with CLI usage (slower read/parse/write on every `/api/chat/all` and `/api/chat/recent`).

Closes #168.

## How

- **`prune(agentName, validSessionIds)`** removes entries whose `sessionId` is absent from the provided set and returns the count removed. It writes only when something was actually removed, mirroring the store's existing batch-write semantics (no write on a steady-state, nothing-to-prune listing).
- **Wired into `getAllSessions()`**, guarded to run **only on a full (unlimited) enumeration**. This is the key correctness point: Phase-2 top-N filtering means a *limited* scan does not enumerate every session, so pruning off a limited set would wrongly delete live entries. When `options.limit` is set, no pruning happens at all.
- **Unions valid sessionIds across every directory that shares a metadata key.** Multiple unattributed directories all write to `adhoc.json`, so pruning per-directory-group (as the issue's sketch suggested) would make one adhoc directory delete another's live entries. The valid set for each key is accumulated across all directories before a single prune per key. The valid set includes *all* transcripts physically present (sidechains and non-enriched sessions too), since their cached metadata is legitimately keyed by sessionId.

## Test coverage

- `session-metadata.test.ts` — `prune()` unit tests: removes absent entries; keeps present ones (fields intact); no-op / no file write when nothing changed (asserts file mtime unchanged); returns 0 and creates no file when none exists; union-across-shared-`adhoc`-key.
- `session-discovery.test.ts` — integration: a full scan prunes each key with the correct union of live sessionIds (adhoc = union of both loose dirs; attributed agent = its own set); **a limited scan (`{ limit: 1 }`) does NOT prune** at all.

All `@herdctl/core` tests pass locally (`build`, `typecheck`, and the two touched suites green; full suite green except one pre-existing root-only env failure in `directory.test.ts` — a read-only-dir permission assertion that root bypasses; unrelated to this change and passes in CI).

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed stale session metadata during complete session scans.
  * Preserved metadata during limited or partial scans to prevent accidental removal.
  * Correctly reconciled metadata shared across multiple session directories.

* **Tests**
  * Added coverage for stale-entry removal, retention of active sessions, missing metadata, no-op pruning, and shared directory scenarios.

### DIFF
--- a/packages/core/src/state/session-discovery.ts
+++ b/packages/core/src/state/session-discovery.ts
@@ -895,7 +895,8 @@ export class SessionDiscoveryService {
     // one adhoc directory would delete another's live entries. Only populated on
     // a full (unlimited) scan; a limited scan does NOT enumerate every session,
     // so pruning off it would wrongly delete live entries.
-    const validSessionIdsByKey = limit === undefined ? new Map<string, Set<string>>() : undefined;
+    const validSessionIdsByKey: Map<string, Set<string>> | undefined =
+      limit === undefined ? new Map<string, Set<string>>() : undefined;
 
     for (const dir of directories) {
       if (validSessionIdsByKey) {
--- a/packages/core/src/state/session-discovery.ts
+++ b/packages/core/src/state/session-discovery.ts
@@ -1045,6 +1045,16 @@ export class SessionDiscoveryService {
     // Runs only on a full scan (validSessionIdsByKey is undefined when a limit
     // was applied). Pruning writes only when it actually removes an entry, so a
     // steady-state listing with no dead sessions performs no extra writes.
+    //
+    // Known limitation: only keys discovered in *this* scan are reconciled. A
+    // directory with zero .jsonl files is skipped in Phase 1, so a key whose
+    // transcripts are ALL gone (e.g. an agent removed from config, orphaning its
+    // <agent>.json) contributes no entry here and is never pruned. This is
+    // mitigated for the primary offender — the shared "adhoc" key is reconciled
+    // as long as any one unattributed directory still has sessions. Fully
+    // reconciling orphaned metadata files would require enumerating
+    // session-metadata/*.json rather than only the keys seen this scan; that's a
+    // larger change left as a follow-up.
     if (validSessionIdsByKey) {
       for (const [metadataKey, validIds] of validSessionIdsByKey) {
         await this.sessionMetadataStore.prune(metadataKey, validIds);
--- a/packages/core/src/state/session-discovery.ts
+++ b/packages/core/src/state/session-discovery.ts
@@ -888,7 +888,30 @@ export class SessionDiscoveryService {
     // Phase 3: Enrich sessions (only selected ones when limit is set)
     const groups: DirectoryGroup[] = [];
 
+    // Accumulate the sessionIds that physically exist on disk for each metadata
+    // key, so stale metadata entries can be reconciled away afterwards (#168).
+    // Multiple directories can share a key — every unattributed directory uses
+    // "adhoc" — so we must union across directories before pruning, otherwise
+    // one adhoc directory would delete another's live entries. Only populated on
+    // a full (unlimited) scan; a limited scan does NOT enumerate every session,
+    // so pruning off it would wrongly delete live entries.
+    const validSessionIdsByKey = limit === undefined ? new Map<string, Set<string>>() : undefined;
+
     for (const dir of directories) {
+      if (validSessionIdsByKey) {
+        let validIds = validSessionIdsByKey.get(dir.metadataKey);
+        if (!validIds) {
+          validIds = new Set<string>();
+          validSessionIdsByKey.set(dir.metadataKey, validIds);
+        }
+        // Every transcript present in this directory is a live session for this
+        // key — include sidechains and non-enriched sessions, all of which carry
+        // legitimately-cached metadata keyed by sessionId.
+        for (const { sessionId } of dir.sessionFiles) {
+          validIds.add(sessionId);
+        }
+      }
+
       const sessions: DiscoveredSession[] = [];
       const autoNameUpdates: Array<{ sessionId: string; autoName?: string; mtime: string }> = [];
       const previewUpdates: Array<{ sessionId: string; preview?: string; mtime: string }> = [];
@@ -1014,6 +1037,16 @@ export class SessionDiscoveryService {
           sessionCount: visibleSessionCount,
           sessions,
         });
+      }
+    }
+
+    // Reconcile metadata against the sessions that still exist on disk (#168).
+    // Runs only on a full scan (validSessionIdsByKey is undefined when a limit
+    // was applied). Pruning writes only when it actually removes an entry, so a
+    // steady-state listing with no dead sessions performs no extra writes.
+    if (validSessionIdsByKey) {
+      for (const [metadataKey, validIds] of validSessionIdsByKey) {
+        await this.sessionMetadataStore.prune(metadataKey, validIds);
       }
     }
 
--- a/.changeset/prune-stale-adhoc-session-metadata.md
+++ b/.changeset/prune-stale-adhoc-session-metadata.md
@@ -1,0 +1,15 @@
+---
+"@herdctl/core": patch
+---
+
+Prune stale session metadata entries during full session scans (#168)
+
+`SessionMetadataStore` accumulated autoName/preview/sidechain/usage cache entries
+for every discovered session but never removed them, so files like `adhoc.json`
+grew unboundedly — entries persisted long after the underlying JSONL transcript
+was deleted. Added `SessionMetadataStore.prune(agentName, validSessionIds)` and
+wired it into `getAllSessions()`, which reconciles each metadata key against the
+sessions still present on disk. Pruning runs only on a full (unlimited)
+enumeration and unions valid sessionIds across every directory sharing a key
+(all unattributed dirs share `"adhoc"`), so a limited/top-N scan never deletes
+live entries, and it writes only when something was actually removed.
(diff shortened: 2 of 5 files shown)


packages/core/src/state/session-discovery.ts
```diff
@@ -1017,6 +1040,16 @@ export class SessionDiscoveryService {
...
 
+    // Reconcile metadata against the sessions that still exist on disk (#168).
+    // Runs only on a full scan (validSessionIdsByKey is undefined when a limit
+    // was applied). Pruning writes only when it actually removes an entry, so a
+    // steady-state listing with no dead sessions performs no extra writes.
+    if (validSessionIdsByKey) {
+      for (const [metadataKey, validIds] of validSessionIdsByKey) {
+        await this.sessionMetadataStore.prune(metadataKey, validIds);
+      }
+    }
```
edspencer, line 1062: **LOW · correctness** — A metadata key whose transcripts are ALL deleted is never pruned

Pruning only runs for `metadataKey`s that appear in `validSessionIdsByKey`, which is built exclusively from directories that survived Phase 1. A directory with zero `.jsonl` files is skipped earlier (`if (sessionFiles.length === 0) continue;`, ~line 790), so it never contributes a key.

Consequence: if *every* transcript for a given key is deleted, that key produces no `DirectoryInfo`, so `prune(key, ...)` is never called and the corresponding `<key>.json` retains all of its now-dead entries forever — the exact unbounded-growth condition the PR sets out to fix.

This is largely mitigated for the main offender (`adhoc`): because all unattributed dirs share the `adhoc` key and it's unioned, as long as *any one* adhoc directory still has sessions, dead entries from fully-emptied adhoc dirs are reconciled away. The gap is real only for a per-agent key (or the whole `adhoc` set) whose transcripts are *all* gone — e.g. an attributed agent all of whose sessions were deleted, or an agent removed from config (its `<agent>.json` becomes orphaned and is never scanned again).

Not blocking — the production case (110 stale `adhoc` entries) is handled well. Flagging as a known limitation: fully reconciling orphaned metadata files would require enumerating existing `session-metadata/*.json` files rather than only the keys discovered this scan, which is a larger design change and reasonably out of scope here. Worth a comment or a follow-up issue.

edspencer (commented):
Adds `SessionMetadataStore.prune(agentName, validSessionIds)` and wires it into `SessionDiscoveryService.getAllSessions()` to reconcile stale session-metadata cache entries against transcripts still on disk.

The core correctness reasoning holds up well. I read `session-metadata.ts` and `session-discovery.ts` in full and traced the valid-set construction. The two dangerous failure modes are both correctly avoided:
- **Limited scans don't prune** (`validSessionIdsByKey` is `undefined` when `limit` is set), so top-N enumeration can't delete live entries.
- **Shared keys are unioned** before a single prune per key, and the valid set deliberately includes *all* transcripts physically present (sidechains, non-enriched, and — for docker/collision dirs — a superset of sessions), so pruning is never a subset of what's live. A superset only means *under*-pruning, which is safe.

`prune()` mirrors the store's existing batch-write semantics (mutates the cached object then writes only when `pruned > 0`), consistent with `batchSetAutoNames`/`batchSetPreviews`. Tests cover the union case, the no-write no-op (mtime assertion), missing-file, and the limited-scan-skips-prune integration path. One low-severity residual-growth gap noted below.

edspencer:
Verified the added comment against the surrounding code: the Phase-1 empty-directory skip (`sessionFiles.length === 0` → `continue`) and the fact that the prune loop only iterates keys present in `validSessionIdsByKey` confirm the documented limitation is accurate. The `adhoc`-key union mitigation and the "enumerate session-metadata/*.json" follow-up suggestion both match the actual code paths. Nothing to flag — the comment matches the code's real behavior.

Reviewed 1 changed file (2 hunks); 0 findings.

edspencer (commented):
This diff adds an explicit type annotation (`Map<string, Set<string>> | undefined`) to the `validSessionIdsByKey` declaration in `getAllSessions()`. It is a purely cosmetic change with no runtime effect — the annotated type exactly matches what TypeScript already infers from the `limit === undefined ? new Map(...) : undefined` ternary, so both branches and all downstream `if (validSessionIdsByKey)` guards behave identically.

packages/core/src/state/session-discovery.ts
edspencer: Good catch — agreed this is a real residual gap and out of scope for the core fix. Documented it as a known limitation with the adhoc mitigation at the prune site in 3d633c6; a full reconciliation (enumerating `session-metadata/*.json`) is left as a follow-up.

edspencer (commented):
This diff is documentation-only: it adds a "Known limitation" comment above the prune loop in `getAllSessions()` (plus the explicit type annotation on `validSessionIdsByKey`). No runtime behavior changes. The comment accurately describes the residual gap — a metadata key whose transcripts are all gone (e.g. an agent removed from config) is skipped in Phase 1 and thus never pruned — and correctly notes the mitigation for the primary `adhoc` case and the follow-up path (enumerating `session-metadata/*.json`).

PR #400, opened 2026-07-22
fix(core): emit in-flight tool_use blocks so running tools survive transcript rehydration (#399)

### LINKED ISSUE #399: parseSessionMessages drops in-flight (unpaired) tool_use blocks — running tool calls vanish on transcript rehydration
## Summary

`parseSessionMessages` (`packages/core/src/state/jsonl-parser.ts`) never emits a `tool_use` that has no matching `tool_result` yet — i.e. **a tool call that is still in flight**. The block is stashed in `pendingToolUses` and only becomes a `role:"tool"` `ChatMessage` once its `tool_result` arrives. A tool that never completes (because the turn is mid-flight when the transcript is read) stays stuck in the map and is silently dropped from the reconstructed history.

This makes a REST/transcript rehydration structurally unable to represent a running tool call. It's the history-side twin of the live-stream gap tracked in **paddock#175**.

## Impact (observed in Paddock)

Paddock renders a chat two ways: a live WebSocket stream (shows in-flight tool activity fine) and a history rebuild from the transcript via `parseSessionMessages` (used on every page load/refresh). While a long-running foreground `Agent`/Task sub-agent is executing:

- **Live view:** the `Agent` tool block shows `RUNNING` with its sub-agent's tool calls streaming in.
- **After a refresh:** the `Agent` block **vanishes** — the rebuilt history omits the unpaired `tool_use`, so the last visible entry is the assistant *text* that preceded it. It reappears only once the sub-agent finishes and the `tool_result` lands.

Reproduced against a real transcript: tail is `assistant text → assistant tool_use:Agent → (later) user tool_result`; with the final line absent (mid-turn), the tool block is not emitted.

## Root cause

In `parseSessionMessages`:

- Assistant `tool_use` blocks are captured into `pendingToolUses` (jsonl-parser.ts:~360) but **not emitted**.
- A `role:"tool"` `ChatMessage` is pushed only in the `tool_result` branch (jsonl-parser.ts:~282), which then deletes the pending entry.
- Any `tool_use` left in `pendingToolUses` at EOF is never surfaced.

## Proposed fix (option 1 from the diagnosis)

Emit in-flight `tool_use` blocks as `role:"tool"` messages in a **pending/running** state, upgraded in place when the matching `tool_result` arrives.

Suggested shape:

- Add an optional discriminator to `ChatToolCall`, e.g. `pending?: boolean` (or `status?: "running" | "complete"`), so consumers can render a spinner instead of a completed block. For a pending tool: `output: ""`, `isError: false`, `durationMs` undefined, `pending: true`.
- Prefer **emit-on-`tool_use`, upgrade-on-`tool_result`** over a final EOF flush, so ordering is preserved (the pending block sits in its correct chronological position) and there's a single code path. The `tool_result` branch then updates the already-emitted message (keyed by the tool_use `uuid`) rather than pushing a new one.
- Keep uuid anchoring identical to today (the originating `tool_use` entry's `uuid`) so per-message UI state stays stable across the pending→complete transition and across reloads.
- Respect the existing `limit` option.
- Keep behavior consistent across both loops in the file if the metadata/limited path also pairs tools.

## Tests

Add to `packages/core/src/state/__tests__/jsonl-parser.test.ts`:

- A transcript ending in an unpaired `tool_use` emits exactly one `role:"tool"` message with `pending:true` (empty output, no duration).
- Appending the matching `tool_result` upgrades the same message (same `uuid`) to a completed tool call (`pending` falsy, output/duration populated) — no duplicate.
- Ordering preserved when a pending tool_use is followed by later assistant text.
- `limit` still respected.

## Downstream

Paddock will consume the `pending` flag to render a running tool block on refresh (fixes the disappearing `Agent` box). Cross-ref **paddock#175** (live-stream side). No change needed to the live translator for this issue — this is purely the transcript-parse path.

edspencer:
## Summary

Closes #399.

`parseSessionMessages` (`packages/core/src/state/jsonl-parser.ts`) never emitted a `tool_use` block that had no matching `tool_result` yet — an **in-flight tool call**. The block was stashed in `pendingToolUses` and only became a `role:"tool"` message once its result arrived, so a running tool (e.g. a foreground `Agent`/Task sub-agent) **vanished from the rebuilt history on page refresh** and only reappeared once the sub-agent finished. This is the history-side twin of paddock#175.

## Fix (option 1 from the diagnosis)

**Emit-on-`tool_use`, upgrade-on-`tool_result`** — single code path, chronological order preserved:

- A `tool_use` is emitted immediately as a `role:"tool"` message in a **pending** state: `pending: true`, `output: ""`, `isError: false`, `durationMs` undefined.
- When the matching `tool_result` arrives, the **same message is upgraded in place** (keyed by the tool_use's recorded index) — its `uuid` and array position are preserved, `pending` is dropped, and output/duration/error are filled in. No duplicate entry.
- Pending tool messages are emitted **after** the assistant text of the same line, so a response carrying both text and a tool_use keeps text-before-tool order.
- The `limit` option is respected (a pending message now occupies a slot); an **orphan** `tool_result` with no seen/emitted `tool_use` still falls back to pushing a fresh completed message.

### API change

Adds an optional discriminator `pending?: boolean` to the `ChatToolCall` interface (additive, non-breaking) so consumers can render a running spinner. Chose `pending?: boolean` over `status` per the issue's guidance.

The metadata/limited path (`extractSessionMetadata`) does **not** pair tools, so no change was needed there — the only tool-pairing loop is inside `parseSessionMessages`.

## Tests

Added to `packages/core/src/state/__tests__/jsonl-parser.test.ts` (with 3 new fixtures):

- Unpaired `tool_use` → exactly one `pending: true` tool message (empty output, no duration, anchored to the tool_use uuid).
- Appending the `tool_result` upgrades the **same** message (same uuid) to completed — no duplicate.
- Ordering preserved when a pending `tool_use` precedes later assistant text.
- `limit` still respected with pending tool messages.

## Verification

- `pnpm --filter @herdctl/core test` — all parser tests green (54/54 in the file; the only suite-wide failure is a pre-existing `directory.test.ts` permission test that fails **only** because this dev box runs as root and ignores `chmod 0o444` — it fails identically on untouched `main` and passes in CI's non-root runner).
- `pnpm typecheck` — green across all 11 packages.
- `pnpm build` — green across all 7 packages.
- Changeset added (`@herdctl/core` minor).

## Summary by CodeRabbit

- **New Features**
  - In-progress tool calls now appear in reconstructed conversation history immediately.
  - Pending tool calls can be identified for displaying a spinner or other active-state indicator.
  - Pending entries update to completed results without duplication or losing their position.

- **Bug Fixes**
  - Prevented running tools from disappearing after refreshing or reloading session history.


edspencer (commented):
This PR fixes #399: `parseSessionMessages` now emits an in-flight `tool_use` (one with no `tool_result` yet) immediately as a `role:"tool"` message in a `pending` state, and upgrades that same message in place when the `tool_result` later arrives — so a running tool survives transcript rehydration instead of vanishing. Adds an additive `pending?: boolean` to `ChatToolCall`.

I read `jsonl-parser.ts` in full, the `tool-parsing.ts` extraction helpers, the new/updated tests and all three new fixtures, and traced the two consumers of `parseSessionMessages` (`session-discovery.ts`'s mtime-keyed message cache, and the test suite). I hand-traced the emit/upgrade/limit paths against the `multi-tool`, `split-tool-blocks`, and the three new pending fixtures. The logic holds up: `messages` is append-only so the recorded `messageIndex` never shifts, chronological order (text-before-tool) is preserved, the `limit` cap is honored on both the pending-emit and orphan-result paths, and a completed call correctly drops `pending`. Could not run the suite (deps not installed / offline), but the assertions match my trace. No blocking issues found.

edspencer:
Key checks performed:
- **Index stability**: `pending.messageIndex` is used to upgrade in place; `messages` is only ever `push`ed (never spliced), so recorded indices stay valid. ✓
- **Ordering**: pending tool messages are emitted after the assistant text of the same line, and completed tools are upgraded in their original (tool_use) position rather than re-pushed at result time — order matches existing `multi-tool`/`split-tool-blocks` expectations. ✓
- **limit**: honored at loop top, on assistant-text emit, on each pending-emit (`break`), and on the orphan-result fallback. When `limit` suppresses the pending emit, `messageIndex` stays undefined and the later result falls to the limit-guarded orphan path. ✓
- **Cache interaction**: `session-discovery.ts` caches parsed messages keyed by file mtime; when the tool_result is appended live the mtime changes and the reparse upgrades the pending entry, so no stale-pending bug. ✓
- **Unidentified blocks**: `tool_use` blocks with no `id` are skipped for emission (can't be paired). ✓

No findings to report — reads as a correct, well-tested, additive change.

Reviewed 6 changed files (12 hunks); 0 findings.

Issue #148, opened 2026-02-25
encodedPath derivation is lossy — directory paths with hyphens can collide


edspencer:
## Bug

In `packages/web/src/server/routes/chat.ts`, `encodedPath` is derived via:

```typescript
session.workingDirectory.replace(/[/\\]/g, "-")
```

This is a non-invertible transformation. Directories containing hyphens collide:

- `/home/user/my-project` → `-home-user-my-project`
- `/home/user/my/project` → `-home-user-my-project`

Since `encodedPath` is used for routing (`/chats/:encodedPath/:sessionId`) and server-side lookup (`groups.find(g => g.encodedPath === encodedPath)`), collisions can silently route users to the wrong directory group's sessions.

Hyphens in directory names are extremely common (npm projects, GitHub repos), making this a realistic collision scenario.

## Affected Paths

- `toRecentFrontendSession()` in `chat.ts` — generates encodedPath
- `GET /api/chat/all/:encodedPath` — looks up group by encodedPath
- `GET /api/chat/sessions/by-path/:encodedPath/:sessionId` — fetches session by encodedPath
- Frontend route `/chats/:encodedPath/:sessionId`
- `encodePathForCli()` in core uses the same scheme

## Fix Direction

Replace the lossy encoding with a reversible, URL-safe scheme such as `encodeURIComponent(workingDirectory)` or base64url encoding. Update both the encoder and all consumers/decoders.

## Files

- `packages/web/src/server/routes/chat.ts` ~line 78
- `packages/core/src/state/session-discovery.ts` (if `encodePathForCli` exists there)
- Frontend routing in `App.tsx`

## Origin

PR #144 CodeRabbit review

edspencer:
Fixed in **@herdctl/core 5.13.1** (PR #265). The lossy encoder must stay byte-compatible with Claude Code (verified empirically), so colliding encoded dirs are now disambiguated at the session level using the `cwd` recorded in each transcript — only when a collision is actually detected (zero extra I/O otherwise). Plus a @herdctl/web routes encoding fix. Tests added.

PR #402, opened 2026-07-22
fix(core): make getAgentSessions worktree-aware (#401)

### LINKED ISSUE #401: getAgentSessions is not worktree-aware — sessions that enter a native git worktree drop out of discovery
## Summary

`getAgentSessions` discovers an agent's sessions from a **single** transcript bucket derived from the agent's registered working directory:

```ts
// packages/core/src/state/session-discovery.ts
const sessionDir = path.join(this.claudeHomePath, "projects", encodePathForCli(workingDirectory));
```

But Claude Code's **native git-worktree support** deliberately relocates a session's transcript when the agent enters a worktree. Per the [worktree docs](https://code.claude.com/docs/en/worktrees) (behaviour gated to Claude Code ≥ 2.1.198):

> "When Claude enters or exits a worktree that Claude Code created with git, the transcript follows: Claude Code records the session under the session's new working directory, the same way `/cd` does, so `/desktop` and `--resume` find it there."

A worktree lives at `<workingDir>/.claude/worktrees/<name>` (native default), which encodes to a **different** `~/.claude/projects/*` bucket than the agent's registered `workingDirectory`. So once an agent enters a worktree, its session is no longer enumerated by `getAgentSessions` — it silently drops out of discovery/attribution even though the transcript is intact.

This is intended Claude Code behaviour, not a bug in the CLI/SDK — it's present and active in the version that actually runs session-mode keepers here (**Agent SDK 0.3.216 → bundled Claude Code 2.1.216**, past all the recent worktree fixes). The gap is that herdctl's discovery predates worktree support and assumes one fixed bucket per agent.

## Observed impact (Paddock)

A spawned keeper (Paddock repo-backed `herdctl` project) was told to "work in a worktree." It used the native `EnterWorktree` tool (no manual `git worktree add`), creating `.claude/worktrees/fix-399-inflight-tooluse`. Its transcript relocated to the worktree bucket:

```
~/.claude/projects/-var-lib-paddock-projects-herdctl-herdctl--claude-worktrees-fix-399-inflight-tooluse/<sessionId>.jsonl
```

Paddock scans only the checkout bucket (`-var-lib-paddock-projects-herdctl-herdctl`), so the chat renders empty in the UI — the primary working chat became invisible while it did the actual work. "Spawn a keeper that uses a worktree" is currently self-defeating for visibility.

## Proposed fix — worktree-aware discovery

In `getAgentSessions`, union the agent's `workingDirectory` bucket with every `~/.claude/projects/*` bucket whose **decoded** path is a descendant of `workingDirectory` (which covers `<workingDir>/.claude/worktrees/*` and any subdir the agent `cd`s into). Reuse the existing directory mtime-cache to bound the extra `readdir` cost.

- Attribution: worktree sessions must still attribute to the owning agent. The attribution index + set-difference discovery key on session id, so a union should be safe — please add a test asserting a worktree-bucket session lists under the correct agent and isn't double-counted.
- Sort/limit semantics (mtime-desc, top-N enrichment) should hold across the unioned set.
- Docker path unchanged (host `docker-sessions/` dir).

## Explicitly do NOT

- Do **not** override/pin the session's resume cwd back to the checkout to keep the bucket stable. That fights documented SDK behaviour — Claude Code deliberately returns a resumed session to its worktree (and falls back to the launch dir only if the worktree is gone). Discovery should follow the SDK's model, not fight it.

## Acceptance

- A session that enters a native worktree under its agent's working dir remains listed by `getAgentSessions` for that agent (regression test with a fixture bucket at `<workingDir>/.claude/worktrees/<name>`).
- No duplicate listing when the same session appears only in the worktree bucket.

## Context

Downstream of Paddock's spawned-keeper flow; the invisible chat was the implementation session for #399. Related: Paddock will consume this automatically via `getAgentSessions` (no Paddock change needed once herdctl ships).

edspencer:
Closes #401.

## Problem

`SessionDiscoveryService.getAgentSessions` discovered sessions from a **single** `~/.claude/projects/{encoded-workingDir}` bucket. Claude Code's native git-worktree support (≥ 2.1.198, active on the bundled 2.1.216) deliberately **relocates** a session's transcript to the worktree's cwd bucket (`<workingDir>/.claude/worktrees/<name>`) when the agent enters a worktree via `EnterWorktree`. That bucket encodes to a *different* directory, so a session that entered a worktree silently dropped out of discovery/attribution even though its transcript was intact — the chat rendered empty downstream (e.g. in Paddock).

## Fix

`getAgentSessions` is now worktree-aware:

- **Union of buckets** — new private `collectWorktreeAwareBuckets(workingDirectory)` returns the agent's own bucket plus every `~/.claude/projects/*` bucket whose **decoded** path is a strict descendant of the working directory (covers `.claude/worktrees/*` and any subdir the agent `cd`s into). Descendant test is on the decoded path via `path.resolve(...).startsWith(workingDir + sep)`.
- **Per-bucket working directory** — each session is tagged with its bucket's working directory, threaded into enrichment (`resolveSidechain`/`resolveAutoName`/`resolvePreview` read the correct transcript file) and reported as the session's `workingDirectory`, so downstream `getSessionMessages` resolves the right bucket. The decoded bucket path re-encodes back to the same directory name (encode/decode round-trips for `[A-Za-z0-9-]` names).
- **Dedupe + sort/limit preserved** — the union is deduped by session id (keeping the newest mtime — guards a lingering enter/exit copy so there's no double-listing), re-sorted mtime-descending, and the top-N `limit` enrichment is applied across the whole merged set.
- **Attribution unchanged as the gate** — the attribution index keys on session id, so worktree-bucket sessions map to the owning agent, and an over-included bucket (e.g. a lossy-encoding false positive, issue #148) contributes nothing spurious.
- **Perf** — per-bucket listings reuse the existing mtime-cache-backed `listSessionFiles`; the one extra cost is a single `readdir` of `~/.claude/projects/` per call (same as `getAllSessions` already does).
- **Docker path unchanged** — Docker agents still scan the single flat host `docker-sessions/` dir; descendant buckets are not scanned for them.

Per the issue's explicit constraint, this does **not** pin the resumed session's cwd back to the checkout to keep the bucket stable — it follows the SDK's documented model (Claude Code intentionally returns a resumed session to its worktree) rather than fighting it.

## Tests

Six new cases under `getAgentSessions > worktree-aware discovery`:
- lists a session that relocated into a native worktree bucket (alongside the checkout bucket)
- lists a worktree-only session exactly once (no duplicate; reports the worktree path; resumable)
- sorts across the unioned buckets and honours the top-N `limit`
- does not union an unrelated project's bucket
- still gates worktree-bucket sessions through attribution (different-agent session excluded)
- does not scan worktree buckets for Docker agents

Existing single-bucket behaviour is unregressed.

## Verification

- `pnpm --filter @herdctl/core typecheck` — pass
- `pnpm --filter @herdctl/core build` — pass
- `pnpm --filter @herdctl/core test` — **3406 pass / 29 skipped**, 78 in `session-discovery.test.ts`. The one failure (`directory.test.ts > throws StateDirectoryCreateError when parent directory is not writable`) is a **pre-existing, environment-only** artifact: the test `chmod 0o444`s a parent then expects a write to reject, which uid 0 (root, the CI/dev-box user) bypasses. It is unrelated to this change (touches `directory.ts`, not session discovery).
- Changeset added (`@herdctl/core`, **minor**).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session discovery for projects using Git worktrees.
  * Sessions from relevant worktree directories now appear alongside primary project sessions.
  * Results are deduplicated, sorted by most recent activity, and limited consistently across all discovered sessions.
  * Unrelated project sessions remain excluded, and Docker session behavior is unchanged.

### DIFF
--- a/packages/core/src/state/session-discovery.ts
+++ b/packages/core/src/state/session-discovery.ts
@@ -503,9 +503,90 @@ export class SessionDiscoveryService {
   }
 
   /**
+   * Collect the transcript buckets to scan for a non-Docker agent.
+   *
+   * Always includes the agent's own `~/.claude/projects/{encoded-workingDir}`
+   * bucket, then unions in every other `~/.claude/projects/*` bucket whose
+   * **decoded** path is a strict descendant of `workingDirectory`.
+   *
+   * This is what makes discovery worktree-aware: Claude Code's native
+   * git-worktree support relocates a session's transcript to the worktree's cwd
+   * bucket when the agent enters a worktree (worktrees live at
+   * `<workingDir>/.claude/worktrees/<name>`, and any subdir the agent `cd`s into
+   * is likewise a descendant). Those buckets encode to different directory names,
+   * so a single-bucket scan silently drops the session. Unioning descendant
+   * buckets keeps it discoverable.
+   *
+   * Safety: the caller still gates every session through the attribution index
+   * (keyed on session id), so an over-included bucket — e.g. a lossy-encoding
+   * false positive from {@link encodePathForCli} (issue #148) — contributes no
+   * spuriously-attributed sessions. The per-bucket listing goes through the
+   * mtime-cache-backed {@link listSessionFiles}, bounding the extra readdir cost.
+   *
+   * Each returned bucket carries the working directory to use for that bucket's
+   * transcript reads and to report as the session's `workingDirectory`. For a
+   * descendant bucket this is the bucket's decoded path, which re-encodes back to
+   * the same directory name (the encode/decode round-trips for any name composed
+   * only of `[A-Za-z0-9-]`, as bucket names are).
+   *
+   * @param workingDirectory - The agent's registered working directory
+   * @returns The primary bucket plus any descendant (worktree/subdir) buckets
+   */
+  private async collectWorktreeAwareBuckets(
+    workingDirectory: string,
+  ): Promise<Array<{ sessionDir: string; workingDirectory: string }>> {
+    const projectsDir = path.join(this.claudeHomePath, "projects");
+    const primaryEncoded = encodePathForCli(workingDirectory);
+
+    // The primary bucket is always scanned, even if it doesn't exist yet
+    // (listSessionFiles handles ENOENT).
+    const buckets: Array<{ sessionDir: string; workingDirectory: string }> = [
+      { sessionDir: path.join(projectsDir, primaryEncoded), workingDirectory },
+    ];
+
+    let encodedPaths: string[];
+    try {
+      encodedPaths = await readdir(projectsDir);
+    } catch (error) {
+      if ((error as NodeJS.ErrnoException).code !== "ENOENT") {
+        logger.warn(
+          `Failed to read projects directory for worktree discovery: ${projectsDir}: ${(error as Error).message}`,
+        );
+      }
+      return buckets;
+    }
+
+    const resolvedWorkingDir = path.resolve(workingDirectory);
+    const descendantPrefix = resolvedWorkingDir + path.sep;
+
+    for (const encodedPath of encodedPaths) {
+      // The primary bucket is already included.
+      if (encodedPath === primaryEncoded) {
+        continue;
+      }
+
+      // Decode the bucket name back to a path and keep it only if it's a strict
+      // descendant of the working directory. `path.resolve` normalises the
+      // decoded path (collapsing e.g. the `//` that `.claude` decodes to).
+      const resolvedDecoded = path.resolve(decodePathForDisplay(encodedPath));
+      if (!resolvedDecoded.startsWith(descendantPrefix)) {
+        continue;
+      }
+
+      buckets.push({
+        sessionDir: path.join(projectsDir, encodedPath),
+        workingDirectory: decodePathForDisplay(encodedPath),
+      });
+    }
+
+    return buckets;
+  }
+
+  /**
    * Get sessions for a specific agent.
    *
-   * Returns sessions from the agent's working directory, attributed and
+   * Returns sessions from the agent's working directory (plus any native
+   * git-worktree / subdirectory buckets nested under it), attributed and
    * enriched with custom names from the metadata store.
    *
    * @param agentName - The agent's qualified name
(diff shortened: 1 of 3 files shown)


edspencer (commented):
This PR makes `SessionDiscoveryService.getAgentSessions` worktree-aware: it unions the agent's primary `~/.claude/projects/<encoded>` bucket with every projects bucket whose decoded path is a strict descendant of the working directory (covering Claude Code's native git-worktree relocation), dedupes by session id keeping the newest mtime, re-sorts mtime-desc, applies the top-N limit across the merged set, and tags each session with its bucket's working directory so enrichment/`getSessionMessages` resolve the right transcript. Docker agents keep their single flat bucket.

What I checked: read `collectWorktreeAwareBuckets` and the rewritten `getAgentSessions` in full; verified `encodePathForCli`/`decodePathForDisplay` round-trip for `[A-Za-z0-9-]` bucket names (so the decoded worktree path re-encodes to the same directory), and that attribution-by-session-id still gates over-included buckets (the #148 lossy-encoding false-positive case) so nothing is cross-attributed; confirmed the descendant test and dedupe/sort/limit logic. The worktree change itself looks correct and well-tested.

One concern outside the stated scope: this PR also deletes the issue-#168 stale-metadata reconciliation from `getAllSessions`, which is not mentioned anywhere in the description — see the finding.

packages/core/src/state/session-discovery.ts
```diff
@@ -1041,26 +1138,6 @@ export class SessionDiscoveryService {
...
-    // reconciling orphaned metadata files would require enumerating
-    // session-metadata/*.json rather than only the keys seen this scan; that's a
-    // larger change left as a follow-up.
-    if (validSessionIdsByKey) {
-      for (const [metadataKey, validIds] of validSessionIdsByKey) {
-        await this.sessionMetadataStore.prune(metadataKey, validIds);
-      }
-    }
-
     // Sort groups by most recent session mtime descending
```
edspencer, line 1141: **HIGH · bug** — PR silently removes the issue-#168 stale-metadata reconciliation from getAllSessions

This PR is described entirely as making `getAgentSessions` worktree-aware, but it also deletes the issue-#168 stale-metadata reconciliation from `getAllSessions`: the `validSessionIdsByKey` accumulation and the trailing `for (const [metadataKey, validIds] of validSessionIdsByKey) { await this.sessionMetadataStore.prune(...) }` loop are both gone (and the `mockPrune`/prune describe block were removed from the test suite).

Evidence:
- The removed block used to sit right before this `groups.sort(...)` call; there is no `prune` call anywhere in `getAllSessions` now.
- `grep '\.prune('` across `packages/core/src` shows the only remaining references are in `session-metadata.test.ts`. `SessionMetadataStore.prune` (session-metadata.ts:594, doc'd as the #168 reconciler) now has **zero production callers** — it is dead code.

Impact: the exact bug #168 fixed returns — stale session-metadata entries (e.g. the shared `adhoc` key) are no longer reconciled on a full scan, so they accumulate unbounded. Nothing about the worktree union in `getAgentSessions` requires touching `getAllSessions`, so this looks like an accidental revert (e.g. a bad rebase dropping the #168 commit) rather than an intended change.

Please confirm whether this removal is intentional. If not, restore the `validSessionIdsByKey` accumulation in the per-directory loop and the prune reconciliation loop (plus the #168 tests). If it is intentional, it needs its own changeset entry, a rationale, and removal of the now-orphaned `prune` method/tests.

edspencer:
- `collectWorktreeAwareBuckets` (new): primary bucket + descendant buckets via `path.resolve(decoded).startsWith(resolvedWorkingDir + sep)`. Over-inclusion is bounded and gated by attribution; verified the encode/decode round-trip so re-encoding a descendant's decoded path recovers the same bucket for ≤200-char encodings.
- `getAgentSessions` rewrite: merge across buckets → dedupe by id (newest mtime) → sort mtime-desc → top-N limit → enrich with per-bucket `sessionWorkingDir`. Logic is sound; the six new tests exercise relocation, dedupe, cross-bucket sort+limit, unrelated-project exclusion, attribution gating, and the Docker path.
- `getAllSessions`: the `#168` `validSessionIdsByKey` accumulation and the `sessionMetadataStore.prune(...)` reconciliation loop were removed (and the corresponding tests dropped). `SessionMetadataStore.prune` now has zero production callers. This appears unrelated to worktree-awareness and undocumented — flagged below.

Reviewed 3 changed files (13 hunks); ran the verify pass; 1 finding.

PR #404, opened 2026-07-22
fix(core): guard openChatSession resume against a still-live session (#403)

### LINKED ISSUE #403: openChatSession(resume) spawns a second subprocess for an already-live session → SDK self-interrupt (double-resume class)
## Summary

`openChatSession(resume)` spawns a **new `claude` subprocess for a session id that already has a live subprocess**, causing two processes to resume one session → the SDK interrupts the in-flight turn (`[Request interrupted by user]`). This is the fundamental cause of a class of double-resume bugs in downstream consumers (Paddock's keeper recovery, queued-drain, wake, manual Continue).

## Root cause

`openChatSession` / `chatSession(resume)` (→ `job-control.ts:395,409`) unconditionally spawns a fresh subprocess resuming the session id from disk — it does **not** check whether a subprocess for that session is still live. Meanwhile `SessionReaper` deliberately keeps a subprocess alive across the turn boundary (keepAlive + ~15s re-invocation grace; `session-reaper.ts:37,156`), during which `reaper.isSessionLive(id) === true`. A resume issued in that window collides with the live subprocess.

Notably, the **wake scheduler already guards on `isSessionLive`** before re-invoking — `openChatSession` is the one resume path that lacks the equivalent guard, so it's the odd one out.

## Impact

Any consumer that may resume a session while the reaper still holds it live gets a spurious `[Request interrupted by user]` and an aborted turn. Confirmed downstream in Paddock keeper auto-recovery: the recovery re-drive fires while the reaper-kept-alive subprocess is still up → SDK self-interrupt (see paddock spike; real transcript `164c7f89…` @ 21:24). A Paddock-side guard can mitigate its own path, but every resume path (Continue, queued-drain, wake, recovery) is exposed until this is fixed at the source.

## Fix

Before spawning, have `openChatSession`/`chatSession(resume)` consult `SessionReaper.isSessionLive(id)` (as the wake scheduler already does) and either attach to / await the live subprocess, or defer-and-retry the resume until it's reaped — rather than spawning a second competing `claude`. This closes the whole double-resume-interrupt class at the source.

## Tests

`session-reaper.test.ts` already covers `isSessionLive` staying true through keepAlive+grace. Add a resume-collision test: with a session live per the reaper, `openChatSession(resume)` must not spawn a second concurrent subprocess (attaches / defers instead).

## Related

Underlying: herdctl#374 (background task killed at turn boundary). Pairs with the Paddock-side RecoveryEngine `isSessionLive` + defer-and-retry fix.

edspencer:
Closes #403.

## Problem

`openChatSession(resume)` (→ `job-control.ts`) unconditionally spawned a **fresh `claude` subprocess** resuming a session id from disk — even when a subprocess for that session was **still live**. The `SessionReaper` deliberately keeps a subprocess alive across the turn boundary when a turn ended with running background work (keepAlive + the ~15s re-invocation grace, `session-reaper.ts`), during which `reaper.isSessionLive(id) === true`. A resume issued in that window launches a **second** `claude` on the same session id, and the SDK resolves the collision by interrupting the in-flight turn (`[Request interrupted by user]`, emitted by the bundled `claude` binary itself).

This is the fundamental cause of a double-resume-interrupt class biting every downstream resume path: keeper auto-recovery, manual Continue, queued-message drain, and any wake that doesn't already guard.

The **wake scheduler already guards** on `isSessionLive` before firing (`wake-registry.ts:149`). `openChatSession` was the one resume path lacking the equivalent check — the odd one out.

## Fix — defer-and-retry

Before spawning, `openChatSession` now consults `SessionReaper.isSessionLive(id)` and, when the target session is still live, **defers the resume** until the reaper reaps it (idle), then spawns exactly as a normal fresh resume would.

**Why defer rather than attach to the live handle?** A `RuntimeSession`'s `messages` is a single SDK `query` generator — a single-consumer stream, and the original opener may still be iterating it during keepAlive/grace. Handing the same live session back to a second `openChatSession` caller would make two consumers race one generator, and re-`manage()` a session already managed. Deferring preserves the existing contract exactly (a fresh, single-consumer `RuntimeSession`, one subprocess) and directly kills the collision by never spawning the second `claude` while the first is live. Deferring *through* legitimately-running background work is also the desired behavior — you don't want a competing resume interrupting it.

Mechanics:
- New event-driven `SessionReaper.whenSessionReaped(sessionId)` resolves immediately if the id isn't live, else when it's reaped/detached (drained in `unregisterLive`). No polling.
- A bounded ceiling (`ChatSessionOptions.resumeDeferTimeoutMs`, default 5 min) keeps a leaked/never-reaped session from hanging the caller — on ceiling it logs and spawns anyway, which is no worse than the pre-#403 behavior of always spawning immediately.

### No regressions
- **(a) genuinely-idle resume still spawns** — the guard only triggers when `isSessionLive(id)` is true; a fresh session (no `sessionId`) is never deferred.
- **(b) wake path not double-guarded into a deadlock** — `dispatchDue` pre-filters live sessions, so `fire()` reaches `openChatSession` with a non-live id → guard is a no-op. The wake itself doesn't hold the session live, so there's nothing to deadlock on.
- **(c) fork resumes unaffected** — `openChatSession` has no fork/`--fork-session` path, and a fork intentionally starts a new session id that never targets the live one.

## Tests
- `session-reaper.test.ts`: `whenSessionReaped` resolves immediately when not live, on reap, and on detach (existing `isSessionLive`-through-keepAlive+grace coverage kept green).
- New `open-chat-session-resume-collision.test.ts`: with a session live per the reaper, `openChatSession(resume)` spawns **no** second subprocess (mocked SDK `query` asserted uncalled); once the reaper releases (`isSessionLive`→false) the deferred resume proceeds **exactly once**. Plus: a fresh (`resume: null`) session spawns immediately even while another session is live, and the ceiling-elapsed fallback spawns anyway.

Full `@herdctl/core` suite + typecheck + build green (one unrelated pre-existing failure: `state/directory.test.ts`'s read-only-dir test can't fail-to-write under a root test runner; fails identically on pristine `main`).

## Companion
Pairs with the Paddock-side mitigation in edspencer/paddock#397 (RecoveryEngine `isSessionLive` + defer-and-retry, which guards only its own path). Once #403 lands, #397's reaper-liveness guard becomes belt-and-suspenders — this closes the whole class at the source, for every resume path.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate subprocesses when resuming a session that is already active.
  - Resumed sessions now wait until the existing session ends before restarting.
  - Added a five-minute maximum wait so resumes proceed even if cleanup does not complete.

- **Tests**
  - Added coverage for active-session collisions, fresh sessions, timeout behavior, and session cleanup.

### DIFF
--- a/.changeset/soft-lions-attack.md
+++ b/.changeset/soft-lions-attack.md
@@ -1,0 +1,9 @@
+---
+"@herdctl/core": minor
+---
+
+Fix double-resume interrupt class: `openChatSession` no longer spawns a second `claude` for a session that already has a live subprocess (edspencer/herdctl#403).
+
+The `SessionReaper` deliberately keeps a session's subprocess alive across the turn boundary while background work runs (keepAlive) or during the ~15s re-invocation grace. Resuming that same session id in that window launched a competing `claude`, and the SDK resolved the collision by interrupting the in-flight turn (`[Request interrupted by user]`) — biting every downstream resume path (keeper auto-recovery, manual Continue, queued-message drain, and any unguarded wake).
+
+`openChatSession` now consults `SessionReaper.isSessionLive(id)` before spawning — the same guard the wake registry already applied — and defers a real (non-fork) resume until the session is reaped, then spawns exactly as a normal fresh resume would. A bounded ceiling (`ChatSessionOptions.resumeDeferTimeoutMs`, default 5 min) keeps a leaked/never-reaped session from hanging the caller. Fresh (non-resume) sessions and the already-guarded wake path are unaffected. Adds `SessionReaper.whenSessionReaped(sessionId)`.
--- a/packages/core/src/fleet-manager/__tests__/open-chat-session-resume-collision.test.ts
+++ b/packages/core/src/fleet-manager/__tests__/open-chat-session-resume-collision.test.ts
@@ -1,0 +1,205 @@
+/**
+ * Tests for the #403 resume-collision guard in FleetManager.openChatSession():
+ *
+ *   The SessionReaper deliberately keeps a session's `claude` subprocess alive
+ *   across the turn boundary while it holds background work (keepAlive) or during
+ *   the re-invocation grace. A resume issued in that window must NOT spawn a
+ *   second `claude` on the same session id — two processes resuming one session
+ *   collide and the SDK interrupts the in-flight turn. openChatSession therefore
+ *   defers the resume until the reaper reaps the session, then spawns exactly
+ *   once. A fresh (non-resume) session is never deferred.
+ *
+ * The Claude SDK's `query` is mocked so no real subprocess is spawned; asserting
+ * how many times it was called is how we observe "did/didn't spawn".
+ */
+
+import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("@anthropic-ai/claude-agent-sdk", () => ({
+  query: vi.fn(),
+}));
+
+import { query } from "@anthropic-ai/claude-agent-sdk";
+import type { RuntimeSession } from "../../runner/runtime/interface.js";
+import type { SessionLifecycleSignal } from "../../session/types.js";
+import { FleetManager } from "../fleet-manager.js";
+
+const silentLogger = () => ({
+  debug: vi.fn(),
+  info: vi.fn(),
+  warn: vi.fn(),
+  error: vi.fn(),
+});
+
+const tick = () => new Promise((r) => setTimeout(r, 10));
+
+/** A background task, so a turn_end carrying it decides `keepAlive`. */
+const TASK = { id: "t1", type: "shell", status: "running", description: "server" };
+
+/**
+ * A fake managed RuntimeSession for the reaper to hold — only the members the
+ * reaper touches (close) need to behave; the rest are inert.
+ */
+function fakeManagedSession(): RuntimeSession {
+  async function* empty(): AsyncGenerator<never> {}
+  return {
+    messages: empty(),
+    send: vi.fn().mockResolvedValue(undefined),
+    interrupt: vi.fn().mockResolvedValue(undefined),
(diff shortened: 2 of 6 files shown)


edspencer (commented):
**What this PR does:** Adds a `#403` defer-and-retry guard to `JobControl.openChatSession`. Before spawning a resumed `claude` subprocess, it now consults `SessionReaper.isSessionLive(id)`; if the target session is still held alive (keepAlive / re-invocation grace), it waits event-driven on the new `SessionReaper.whenSessionReaped(id)` until the reaper releases the id, then spawns a normal fresh resume. A bounded ceiling (`ChatSessionOptions.resumeDeferTimeoutMs`, default 5 min) prevents a leaked/never-reaped session from hanging the caller.

**What I checked:** Read the full `openChatSession` body (`job-control.ts` 347-499) including the resume-id resolution precedence, the guard placement, `deferResumeUntilReaped`, and the `raceWithTimeout` helper. Read `session-reaper.ts` `whenSessionReaped`/`unregisterLive` and confirmed both the reap path (`ManagedSessionImpl` line 277) and the detach path (line 354) call `unregisterLive`, which drains and resolves waiters — matching the new tests. Verified `whenSessionReaped` re-checks `liveById` synchronously before enqueuing a resolver (no TOCTOU/missed-wakeup, since JS is single-threaded and there's no `await` between check and push), and the loop re-checks `isSessionLive` each pass so a same-id re-registration is handled. Grepped callers of `openChatSession` (wake fire path in `session-lifecycle-manager.ts` — no self-deadlock since the wake doesn't hold the session live). Confirmed the `lifecycleManager` refactor is behavior-preserving vs. the old `manageLifecycle ? getSessionLifecycle() : undefined`.

**Verdict:** Clean. `raceWithTimeout` handles both settle sides (`void promise.then(done, done)`) so no floating-rejection risk, unref's its timer, and clears it. Guard only triggers for reaper-managed (opt-in) sessions and never for fresh (`resume: null`) sessions. Tests cover defer-then-spawn-once, fresh-not-deferred, and ceiling-elapsed fallback. No findings worth a maintainer's attention.

edspencer:
Reviewed 6 changed files (13 hunks); 0 findings.

Issue #374, opened 2026-07-15
Keeper hangs when a background task is still running at turn-end — SDK kills it, no re-invocation (not the reaper; upstream SDK/CLI)


edspencer:
## Summary

#368 fixed the reap **race** where a background task *completing* reaped the managed session out from under the SDK re-invocation. That fix is confirmed working in the field. But a sibling failure survives it on `@herdctl/core` **5.20.1**: when a background **`Agent` subagent** (as opposed to a background Bash command) is **killed at the keeper turn boundary**, the keeper's re-invocation is dropped and the session stalls — and the kill is stamped `status=killed` / `summary: "…was stopped by user"`, so the UI shows **"stopped by the user" with no user action**.

Reproduced on the **post-fix build** (paddock 0.23.0 / core 5.20.1), so this is **not** #368 and **not** a deploy-restart artifact.

## Relationship to #368

- **#368:** background task **completes** → race reaps session → re-invocation dropped. **Fixed** (deferred-reap grace window).
- **This:** background **`Agent` subagent killed** at turn-end → re-invocation dropped / degenerates to a null turn → session idles and is reaped. **Not covered by the #368 fix.**

Sharp discriminator, observed in the same chat: **background *commands* re-invoke the keeper fine post-fix** (many `Background command "…" completed (exit code 0)` notifications, keeper resumed each time), but background ***Agent* subagents / long pollers get `status=killed`** and the keeper never resumes.

### Repro 1 — post-fix, terminal (paddock keeper, chat `43245224`)

All events 2026-07-15 03:37:57Z → 04:17:35Z UTC — after the 02:41:00Z (0.23.0) and 02:55:09Z (self-mcp) restarts and before the 06:08Z restart ⇒ **build 0.23.0 / core 5.20.1**.

- `04:17:33.795Z` — keeper posts its plan: *"…I'll be re-invoked when it completes, then commit → push → PR → deploy…"*
- `04:17:33.860Z` — foreground turn ends **cleanly**: `stop_hook_summary`, `preventedContinuation:false` (a normal `end_turn`, **not** an interrupt).
- `04:17:35.891/.893Z` — `queue-operation` enqueue+dequeue of a `<task-notification>` for background `Agent` subagent `a7beb89bde2116bf6` ("Install, test, build paddock #175").
- `04:17:35.904Z` — the notification lands: `<status>killed</status>` / `<summary>Agent "Install, test, build paddock #175" was stopped by user</summary>`.
[...]

Issue #368, opened 2026-07-14
Async background-task completion reaps the parent session mid-re-invocation (background_tasks_changed races the re-invoke turn's activity)


edspencer:
## Summary

In session drive-mode, when an **asynchronous background task** (a `run_in_background` Bash, a background `Agent`/`Task`, a `Monitor`) **completes**, the SDK injects a re-invocation turn to hand the parent the task's result. But the session-reaper can reap (close) the session in the same instant, **before that re-invocation turn produces any output** — so the agent is never actually re-invoked and the continuation is silently dropped.

This is the async counterpart of #366 / #367. That fix stopped a *synchronous* subagent (`SubagentStop`) from being mis-mapped to `turn_end`. This bug is a different path: the `background_tasks_changed → 0` branch in the reaper racing the re-invocation turn's `activity` signal. #367 does **not** cover it.

## Impact

The "launch a background task → get re-invoked with its result when it finishes" contract — the core primitive for autonomous cross-turn work in session drive-mode — is **racy and unreliable**. A keeper that kicks off slow work in the background (CI watch, a long build, a deploy, an `npm publish`) and ends its turn saying *"I'll be re-invoked when it resolves"* frequently just **stops**: the background task finishes, the session is reaped mid-re-invocation, and nothing carries the work forward. To the user it looks like the agent silently died mid-task.

## Root cause

`SessionReaper.processSignal` (`packages/core/src/session/session-reaper.ts`):

```js
if (signal.kind === "activity") {
    managed.setAwaitingTasks(false);   // "a new turn is producing output"
    return;
}
if (signal.kind === "background_tasks_changed") {
    if (managed.isAwaitingTasks() && signal.backgroundTasks.length === 0) {
        this.reap(managed, signal.sessionId);   // <-- reaps here
    }
    return;
}
```

`awaitingTasks` is set true on a `turn_end` that decided `keepAlive` (i.e. the turn ended while a background task was still live). The `activity` signal is meant to clear it when the next turn starts producing output. The guard works **only if `activity` is processed before `background_tasks_changed`**.

When an async background task completes, the SDK emits, ~simultaneously:
1. a **re-invocation turn** (the task-completion notification is injected as new turn input) — which emits its `activity` marker once it *starts producing output*, and
2. a **`background_tasks_changed`** system message with the now-empty task list.

[...]

PR #407, opened 2026-07-23
fix(core): defer turn_end reap on resume-with-prompt so a replayed backlog turn can't interrupt the human (#406)

### LINKED ISSUE #406: SessionReaper reaps a replayed backlog turn on resume-with-prompt → aborts + loses the human turn ([Request interrupted by user])
## Summary

In session drive-mode, resuming a **reaped** session **with a human prompt** can self-interrupt and **lose the human's message**. The terminal transcript shows `[Request interrupted by user]` with `interruptedByShutdown: true` beneath a turn that never ran — even though no human hit Stop. Root cause is `SessionReaper` **closing the session on a replayed backlog turn**, which aborts the human turn queued behind it. Proven end-to-end against the real bundled `claude` binary (2.1.216) driving the real `SessionReaper` + `SDKRuntime`.

Distinct from the double-resume class (#403/#404) and from #397: those cover a resume racing a *reaper-live* session (two subprocesses). Here there is **one** subprocess and the session is **not** reaper-live at resume time, so #404's `isSessionLive` guard is correctly skipped — the teardown comes from the reaper reaping the *backlog* turn.

## Mechanism (corrected — the obvious model is wrong)

Two things that are NOT the cause (both disproven on the real binary):
- The CLI does **not** replay persisted `queue-operation` audit-log entries. Injecting synthetic `killed <task-notification>` enqueues and resuming produces zero replay — those log lines are an audit trail; replay is driven by the CLI's background-task state.
- A second user message pushed mid-turn does **not** interrupt — the CLI's async input queue coalesces/appends it gracefully.

What actually happens on `openChatSession(resume)` with a human prompt, when the prior process died mid-turn leaving pending background-task state:
1. The CLI flushes the pending background-task notification as **turn A** (its own turn), *ahead* of the human **turn B**.
2. Turn A ends → Stop hook emits `turn_end` with no background work.
3. `SessionReaper` treats the session as idle and **reaps immediately** (`processSignal` → `reap` → `scheduleClose` → `session.close()`).
4. `session.close()` fires the abort controller while **turn B is in flight** → Claude writes `[Request interrupted by user]` / `interruptedByShutdown: true`, and the human's turn is torn down and lost.

The reaper reaping on the *backlog* turn — before the human turn runs — is the teardown trigger, confirmed by the reaper's own `Reaping idle session … (probe)` log line landing on turn A's `turn_end`.

## Proof (real binary, before/after)

Driving the deployed bundled binary (`2.1.216`) via the SDK through the **real** shipped `SessionReaper` + `SDKRuntime`:
- **BEFORE (current code):** reaper reaps on turn A's `turn_end`; transcript ends `USER(sdk) <task-notification>` (A) → `USER(sdk) <human>` (B) → `USER interruptedByShutdown=true "[Request interrupted by user]"`. Human command never executes (`interrupt=1`, `humanAnswered=false`). Matches a real production session's on-wire signature line-for-line.
- **AFTER (fix):** turn A ends → grace armed → turn B's `activity` cancels the pending reap → human turn runs to completion → session reaped at true idle later (`interrupt=0`, `humanAnswered=true`, no hang).

## Fix

Reaper grace on `turn_end` (mirrors the existing #368 background-drain grace):
- `session/session-reaper.ts`: add `ManageSessionOptions.turnEndReapGraceMs`; on `turn_end` with no background work, if set, `armPendingReap(grace, …)` instead of an immediate `reap()` (an incoming `activity` already cancels a pending reap). Default `0` = today's behavior, unchanged.
- `session/session-lifecycle-manager.ts`: thread it through `manage()`.
- `fleet-manager/job-control.ts` `openChatSession`: on a real (non-fork) resume **with a prompt**, set `turnEndReapGraceMs = DEFAULT_REINVOCATION_GRACE_MS` (15s) so a replayed backlog turn cannot reap the resumed prompt turn; a genuinely final turn's grace still elapses and reaps normally.

A validated candidate implementation exists (reaper-grace + a `hasPendingAsyncQueue()` disk detector that tallies `queue-operation` enqueue−dequeue balance) with unit tests (4 grace + 6 detection) and the real-binary before/after above; full core suite green aside from a pre-existing unrelated `chmod 0o444` test that root bypasses.

## Open design decision (please weigh in)

The candidate **gates** the grace on the disk detector (`enqueue−dequeue balance > 0`). That residue was present in every reproduction (a real production session showed balance 5; a real background-task session showed 1; a clean boundary is 0) — but detection is **correlation, not causation**: the replay is background-task-state-driven and the queue residue is a correlated side-effect, so we can't prove every future replay leaves residue.

**Recommendation: arm the grace unconditionally on any real resume (`sessionId` set), regardless of detection.** It cannot miss, at a cost of ~one reinvocation-grace window (~15s) of extra resident memory per resumed session before it reaps. Given this failure has slipped narrow fixes repeatedly, the belt-and-suspenders is worth the small cost. Detection-gating is available if we want to limit blast radius instead.

## Related

Same family as #374 (background task killed at turn boundary — herdctl closing the session tears down in-flight work). Complements #403/#404 and downstream paddock#397; none of those cover this not-reaper-live backlog-turn reap.

edspencer:
Closes #406.

## The bug

In session drive-mode, a **real (non-fork) resume that carries a human prompt** can
self-interrupt and **lose the human's message**. If the prior process died mid-turn
leaving pending background-task state, the CLI replays that leftover as its **own
turn (turn A)** ahead of the caller's queued prompt **turn B**. Turn A ends with no
background work, so `SessionReaper` reaps **immediately** on its `turn_end`
(`processSignal → reap → scheduleClose → session.close()`), and `close()` aborts the
in-flight human turn — the transcript gets `[Request interrupted by user]` /
`interruptedByShutdown: true` and the human's message is dropped.

This is **not** the #403/#404 double-resume class: the session is **not** reaper-live
at resume, so #404's `isSessionLive` guard is correctly skipped. The trigger is the
reaper reaping the *backlog* turn.

## The fix — reaper grace on `turn_end` (mirrors the #368 background-drain grace)

- **`session/session-reaper.ts`** — add `ManageSessionOptions.turnEndReapGraceMs`. On
  a `turn_end` with no background work, if set, `armPendingReap(grace, …)` instead of
  reaping synchronously; an incoming turn's `activity` cancels the pending reap.
  Default `0` = today's behavior, unchanged.
- **`session/session-lifecycle-manager.ts`** — thread the option through `manage()`.
- **`fleet-manager/job-control.ts` `openChatSession`** — arm the grace
  (`DEFAULT_REINVOCATION_GRACE_MS`, 15s) on a real resume that carries a prompt, so a
  replayed backlog turn can no longer reap the resumed prompt turn. A genuinely final
  turn still reaps once its grace elapses (with an arriving turn's `activity`
  cancelling it). Fresh (non-resume) sessions and `--fork-session` resumes are
  unaffected.

### Design decision: the grace is armed **unconditionally** on resume-with-prompt

Per #406's open-decision section, the grace is armed on **any** real (non-fork)
resume that carries a `sessionId` + prompt, **not gated on disk detection**. The
replay is background-task-state-driven; the on-disk async-input queue residue is only
a *correlated side-effect* (correlation, not causation), so a future replay could
leave no residue. Since this failure has slipped narrow guards repeatedly, the
belt-and-suspenders is worth the cost: ~one reinvocation-grace window (~15s) of extra
resident memory per resumed session before it reaps at true idle.

A `hasPendingAsyncQueue()` / `countPendingAsyncQueueEntries()` detector (net
enqueue−dequeue depth of the CLI async-input queue) is included **as telemetry only**
— logged on resume so production logs can correlate the failure signature. It never
gates reap behavior.

## Merge gate — real-binary before/after on the BUILT artifact

Evidence is **not** "green unit tests." A harness drives the **real deployed bundled
`claude` binary (2.1.216)** through the **compiled `dist/` `SessionReaper` +
`SDKRuntime`** (not ts-source, not a fake claude), against a real OAuth session,
outside the sanitized test env. Turn A (the pending backlog `<task-notification>`) is
injected on-wire with the exact turn structure + reaper signals a real replay
produces (real CLI replay of the backlog-ahead-of-human ordering is timing-flaky —
see the honest note below); turn B is a real human command whose output is verifiable.
The authoritative signature is read from the **on-disk transcript** (the CLI writes
the `[Request interrupted by user]` user turn there even when the SDK consumer stream
ends before yielding it).

**BEFORE (grace = 0, current behavior)** — the reaper reaps on turn A's `turn_end`
and `close()` aborts the human turn. Transcript tail:

```
system
user "Run this exact command and report only its out…"   ← human turn B
assistant [tool_use]
user "\"HUMAN_ANSWER_7f3\""
last-prompt
mode
user <interruptedByShutdown> "[Request interrupted by user]"   ← the signature
=== GRACE=0 | TRANSCRIPT interrupt=1 sawHumanTurn=true | STREAM humanAnswered=false ===
```

Deterministic across repeated runs: `TRANSCRIPT interrupt=1` every time — the exact
`interruptedByShutdown` / `[Request interrupted by user]` signature, human turn torn
down.

**AFTER (grace = 15000, the fix)** — turn A's `turn_end` arms the grace; turn B's
`activity` cancels it; the human turn runs to completion (`RESULT #2 success`); the
session is reaped at true idle ~15s later (grace elapsed, no follow-up turn), no hang.
Transcript tail:

```
system
user "Run this exact command and report only its out…"   ← human turn B
assistant [tool_use]
user "\"HUMAN_ANSWER_7f3\""
assistant "HUMAN_ANSWER_7f3"                              ← human answered
last-prompt
mode
=== GRACE=15000 | TRANSCRIPT interrupt=0 humanExecuted=true sawHumanTurn=true | STREAM humanAnswered=true ===
```

Deterministic across repeated runs: `TRANSCRIPT interrupt=0`, human answered exactly
once, reaped at true idle.

### Fully-real anchor (honest)

Alongside the deterministic injection I also ran fully-real end-to-end replays: a real
background task killed mid-turn leaves un-drained backlog on disk (the shipped
detector reports residue depth 1–2), then the session is resumed with the human as the
initial prompt through the same compiled reaper + runtime — nothing injected. These
confirm faithfulness on three points: **real on-disk residue**, the real `RESULT #1`
no-op resume-ack, and a real reaper **reap-on-`turn_end`**. With the grace on, the
human turn completes and the session reaps at true idle ~15s later (no regression).

### Natural reproduction at production-like residue depth (~5) — attempted, did NOT trigger

To close the gap honestly I pushed the fully-natural path to production-like residue
depth and tried to make the CLI **spontaneously** replay the backlog as a turn ahead
of the human (no injection), on the real 2.1.216 binary through the compiled reaper +
runtime:

- Two backlog mechanisms: (a) many slow `run_in_background` **bash** tasks whose
  completion notifications enqueue un-dequeued (measured residue depth **1, 2, 6**);
  (b) **5 orphaned background Task agents** killed at the boundary — the exact
  production `<status>killed</status>` mechanism (#374) — measured residue depth
  **2, 3, 4**.
- Two resume orderings: human as the initial resume prompt, and no-prompt resume then
  human via `send()` a few seconds later.

**Result: the natural interrupt did NOT reproduce in any trial** (`TRANSCRIPT
interrupt=0` every time, human turn completed). Root cause, observed directly from the
lifecycle-signal trace: at 2.1.216 the CLI delivers the pending backlog **without a
separate turn boundary** — either as a no-op resume-ack (which emits **no** Stop hook /
`turn_end`) or by **coalescing** the backlog notifications *into the same single turn
as the human*. There is exactly **one** `turn_end`, so the reaper reaps **once, after
the human already completed** — it never gets the chance to reap *between* a backlog
turn and the human turn, which is the precise interrupt trigger.

**Honest read:** the specific state where the backlog gets its **own** `turn_end`
ahead of the human (which production transcript `dcd8e17e` exhibited at depth ~5) is
genuinely non-deterministic / version-or-state-dependent and **could not be provoked
on demand** here, even at matching residue depth. This is itself a finding about how
rare/timing-dependent the real trigger is, and is consistent with #406's "real CLI
replay is timing-flaky" note and why prior narrow fixes slipped. The **deterministic
injection harness** — which supplies exactly that turn structure + reaper signals and
reproduces the real `[Request interrupted by user]` transcript signature reliably —
therefore stands as the regression oracle. The fix's mechanism is unconditional and
independent of whether the backlog arrives as its own turn: any real resume-with-prompt
gets the grace, so a backlog turn that *does* get its own `turn_end` can no longer reap
the resumed prompt turn.

## Tests & changeset

- New unit tests: 4 for `turnEndReapGraceMs` grace behavior (synchronous reap when
  `0`; deferred + activity-cancelled; final-turn grace-elapsed reap; re-arm after the
  human turn) and 6 for the async-queue detector.
- Full `@herdctl/core` suite green **except** the one pre-existing, unrelated
  `state/directory.test.ts` case that asserts a non-writable-parent failure — it
  passes as non-root but root bypasses the `chmod 0o444` check. Untouched by this PR.
- Changeset added (`@herdctl/core`, minor).

### DIFF
--- a/packages/core/src/fleet-manager/job-control.ts
+++ b/packages/core/src/fleet-manager/job-control.ts
@@ -12,6 +12,8 @@ import { isAbsolute, join, resolve } from "node:path";
 import type { HookEvent, ResolvedAgent } from "../config/index.js";
 import { type HookContext, HookExecutor } from "../hooks/index.js";
 import {
+  countPendingAsyncQueueEntries,
+  getCliSessionFile,
   JobExecutor,
   RuntimeFactory,
   type RuntimeSession,
@@ -19,6 +21,7 @@ import {
   type SlashCommand,
 } from "../runner/index.js";
 import type { ManagedSession, SessionLifecycleSignal, SessionReaper } from "../session/index.js";
+import { DEFAULT_REINVOCATION_GRACE_MS } from "../session/index.js";
 import { createJob, getJob, getSessionInfo, readJobOutputAll, updateJob } from "../state/index.js";
 import type { JobMetadata } from "../state/schemas/job-metadata.js";
 import type { FleetManagerContext } from "./context.js";
@@ -427,6 +430,50 @@ export class JobControl {
       );
     }
 
+    // #406: a real resume that carries a human prompt can self-interrupt and lose
+    // the human's message. If the prior process died mid-turn leaving pending
+    // background-task state, the CLI replays that leftover as its OWN turn (turn A)
+    // ahead of the caller's queued prompt turn (turn B). Turn A ends with no
+    // background work, so the reaper reaps immediately on its `turn_end` and closes
+    // the session out from under turn B — interrupting it (`[Request interrupted by
+    // user]` / `interruptedByShutdown`) and losing the message. This is NOT the
+    // #403/#404 double-resume class: the session is not reaper-live at resume, so
+    // that guard is correctly skipped; the trigger is the reaper reaping the
+    // *backlog* turn.
+    //
+    // Fix: hand the managed session a `turnEndReapGraceMs` so a `turn_end` reap is
+    // deferred long enough for turn B's `activity` to cancel it (a genuinely final
+    // turn's grace still elapses and reaps). Arm it UNCONDITIONALLY on any real
+    // (non-fork) resume that carries a prompt — `openChatSession` never forks, so
+    // every resume here qualifies. We deliberately do NOT gate on disk detection:
+    // the replay is background-task-state-driven and the queue residue below is
+    // only a correlated side-effect (correlation, not causation), and this failure
+    // has slipped narrow guards repeatedly — so belt-and-suspenders. Cost is ~one
+    // reinvocation-grace window (~15s) of extra RSS per resumed session before it
+    // reaps at true idle. Fresh sessions (no `sessionId`) are unaffected. See #406.
+    let turnEndReapGraceMs = 0;
+    if (sessionId && options?.prompt) {
+      turnEndReapGraceMs = DEFAULT_REINVOCATION_GRACE_MS;
+      // Best-effort telemetry only (NOT a gate): log whether the resumed session
+      // carries the correlated stale async-input residue, so production logs can
+      // confirm the #406 signature. A detection failure never changes behavior.
+      let residue: number | undefined;
+      try {
+        const workspacePath = resolveAgentWorkspacePath(effectiveAgent);
+        residue = await countPendingAsyncQueueEntries(getCliSessionFile(workspacePath, sessionId));
+      } catch (error) {
+        logger.debug(
+          `Pending-async-queue telemetry skipped for ${sessionId}: ${(error as Error).message}`,
+        );
+      }
+      logger.info(
+        `Resuming session ${sessionId} (${agentName}) with a prompt; deferring turn-end ` +
+          `reaps by ${turnEndReapGraceMs}ms so a replayed backlog turn cannot reap the ` +
+          `resumed prompt turn (#406)` +
+          (residue !== undefined ? ` [pending-async-queue residue: ${residue}]` : ""),
+      );
+    }
+
     const runtime = new SDKRuntime();
     logger.info(`Opening streaming chat session for ${agentName} (sdk runtime)`);
 
@@ -452,7 +499,11 @@ export class JobControl {
     });
 
     if (lifecycle) {
-      managed = lifecycle.manage(session, agentName);
+      managed = lifecycle.manage(
+        session,
+        agentName,
+        turnEndReapGraceMs > 0 ? { turnEndReapGraceMs } : undefined,
+      );
     }
 
     return session;
@@ -1047,6 +1098,22 @@ function raceWithTimeout(promise: Promise<void>, ms: number): Promise<void> {
  * @throws {InvalidWorkingDirectoryOverrideError} If the override is provided but
  *   not a non-empty string
  */
+/**
+ * Resolve the absolute workspace path a resumed session's transcript is keyed by.
+ *
+ * Mirrors the cwd the SDK adapter passes to `claude` (`agent.working_directory`,
+ * which may be a string or `{ root }`), falling back to the process cwd when the
+ * agent has none — the same basis Claude Code uses to encode
+ * `~/.claude/projects/<encoded-cwd>/<sessionId>.jsonl`. Used only for #406's
+ * best-effort residue telemetry, so a wrong guess degrades to "no residue logged".
+ */
+function resolveAgentWorkspacePath(agent: ResolvedAgent): string {
+  const wd = agent.working_directory;
+  if (typeof wd === "string" && wd.trim() !== "") return wd;
+  if (wd && typeof wd === "object" && typeof wd.root === "string") return wd.root;
+  return process.cwd();
+}
+
 function applyWorkingDirectoryOverride(
   agent: ResolvedAgent,
   override: string | undefined,
--- a/.changeset/spicy-donkeys-406.md
+++ b/.changeset/spicy-donkeys-406.md
@@ -1,0 +1,20 @@
+---
+"@herdctl/core": minor
+---
+
+Fix a resume self-interrupt that loses the human's turn (#406). On a real
+(non-fork) resume that carries a human prompt, if the prior process died mid-turn
+leaving pending background-task state, the CLI replays that leftover as its own
+turn ahead of the human turn; the `SessionReaper` reaping on the replayed backlog
+turn's `turn_end` closed the session out from under the in-flight human turn,
+producing `[Request interrupted by user]` / `interruptedByShutdown` and dropping
+the human's message.
+
+`SessionReaper.manage` now accepts a `turnEndReapGraceMs` (via
+`ManageSessionOptions`) that defers a `turn_end` reap so an immediately-following
+turn's `activity` can cancel it — mirroring the existing background-drain grace.
+Default `0` preserves current behavior. `openChatSession` arms this grace
+(`DEFAULT_REINVOCATION_GRACE_MS`, 15s) unconditionally on any real resume that
+carries a prompt, so a replayed backlog turn can no longer reap the resumed prompt
+turn; a genuinely final turn still reaps once its grace elapses. Fresh (non-resume)
+sessions and `--fork-session` resumes are unaffected.
--- a/packages/core/src/runner/index.ts
+++ b/packages/core/src/runner/index.ts
@@ -46,11 +46,13 @@ export type {
   SlashCommand,
 } from "./runtime/index.js";
 export {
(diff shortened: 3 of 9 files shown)


packages/core/src/fleet-manager/job-control.ts
```diff
@@ -427,6 +430,50 @@ export class JobControl {
...
+    let turnEndReapGraceMs = 0;
+    if (sessionId && options?.prompt) {
+      turnEndReapGraceMs = DEFAULT_REINVOCATION_GRACE_MS;
+      // Best-effort telemetry only (NOT a gate): log whether the resumed session
+      // carries the correlated stale async-input residue, so production logs can
+      // confirm the #406 signature. A detection failure never changes behavior.
+      let residue: number | undefined;
+      try {
+        const workspacePath = resolveAgentWorkspacePath(effectiveAgent);
+        residue = await countPendingAsyncQueueEntries(getCliSessionFile(workspacePath, sessionId));
```
edspencer, line 463: **LOW · performance** — Telemetry-only transcript read blocks the resume critical path

`countPendingAsyncQueueEntries()` is `await`ed here (line 463) before `runtime.openSession()` (line 491), so every real resume-with-prompt now blocks on reading and line-scanning the **entire** `<sessionId>.jsonl` transcript purely to log a residue count. `pending-async-queue.ts` does `readFile(path, "utf8")` then `raw.split("\n")` over the whole file — for a long-lived agent whose transcript is many MB, that is a real (hundreds-of-ms-plus) added latency on the resume hot path for data the PR itself documents as "telemetry only, NOT a gate."

Since the value only feeds a log line, it doesn't need to block the session from opening. Consider computing it off the critical path (e.g. fire-and-forget the read and log the residue when it resolves) so a slow/large transcript never delays the resume. Functionally correct as-is — hence low — but the cost lands on every resumed chat session, not just the failure case.

edspencer (commented):
Solid, well-tested fix for the #406 resume self-interrupt. Adds a per-session `turnEndReapGraceMs` to `SessionReaper` so a `turn_end` reap is deferred (mirroring the existing #368 background-drain grace) and can be cancelled by the next turn's `activity`; `openChatSession` arms it unconditionally (15s) on any real resume carrying a prompt. Default `0` preserves current behavior, so fresh/fork sessions are unaffected.

I read `session-reaper.ts` in full (the new `processSignal` branch, `armPendingReap`/`cancelPendingReap`, and the `manage()` plumbing), traced the option through `session-lifecycle-manager.ts` and `job-control.ts` `openChatSession`, verified the `countPendingAsyncQueueEntries`/`getCliSessionFile` exports and the `resolveAgentWorkspacePath` mirror of `cli-runtime.ts`'s cwd logic, and reviewed the new unit tests. The grace/cancel/re-arm state machine is correct, the detector is genuinely telemetry-only (wrapped in try/catch, never gates behavior), and the live-session interaction with the #403 guard is consistent with the existing #368 grace. One low-severity note below on the telemetry read sitting on the resume critical path.

edspencer:
Reviewed 9 changed files (18 hunks); ran the verify pass; 1 finding.

PR #428, opened 2026-08-01
fix(core): skip unreadable transcript entries instead of blanking the session listing (#424)

### LINKED ISSUE #424: One unreadable entry in a transcript folder throws and blanks the entire session listing
## Summary

A single unreadable entry inside a Claude Code transcript folder makes session
listing throw outright. `getAgentSessions` loses the whole agent's list;
`getAllSessions` loses **every agent's** list. The caller gets an exception, not
a partial result — good transcripts sitting next to the bad entry are lost.

Reproduced with the simplest case: a **directory** named `<uuid>.jsonl` inside
`~/.claude/projects/<encoded-cwd>/`.

## Why it isn't caught

On Linux, `open(2)` on a directory **succeeds** — only `read(2)` fails. So
`createLineReader` in `state/jsonl-parser.ts` resolves a reader successfully and
its `stream.on("error")` guard never fires. The failure surfaces later, on the
`for await` inside `isSidechainSession`, as a bare libuv error with no JS stack
frames:

```
Error: EISDIR: illegal operation on a directory, read
  { code: "EISDIR", syscall: "read" }
```

`listSessionFiles` cannot screen it out either: it filters on the `.jsonl`
**name**, and `stat()` on a directory succeeds, so the entry is admitted to the
listing with a perfectly valid mtime and size.

## Blast radius

`isSidechainSession` is reached from `SessionDiscoveryService.resolveSidechain`,
which is called unguarded from two places:

- **`getAgentSessions`** — inside `mapWithConcurrency`, which rejects like
  `Promise.all` (`state/utils/concurrency.ts`). One bad entry rejects the entire
  enrichment.
- **`getAllSessions`** — inside a plain sequential loop that is itself nested in
  a loop over *directories*. One bad entry in one directory therefore aborts
  enrichment for **every agent/directory in the listing**.

Also latent on the same path: `resolveAutoName` and `resolvePreview` both go
through `createLineReader` and would throw identically — `resolveSidechain`
simply runs first.

## Repro

```js
const dir = path.join(claudeHome, "projects", encodePathForCli(workingDir));
await fs.mkdir(path.join(dir, `${crypto.randomUUID()}.jsonl`), { recursive: true });
// plus one perfectly good transcript in the same folder
await fleet.getAgentSessions(agentName); // throws EISDIR; the good session is lost
```

## Suggested direction

Treat a per-entry read failure as "skip this entry", not "fail the listing" —
the enrichment for one session should never be able to take down the listing.

There is already a working precedent in this codebase: `listAdoptableSessions`
(added for #423) catches exactly this and skips the entry with an `"unreadable"`
reason, and on the identical fixture it returns the good session cleanly. That
is the only hardened path today; `getAgentSessions`/`getAllSessions` should get
the same treatment.

Worth deciding deliberately whether a skipped entry should be *reported*
(a `skipped`/`errors` channel on the result) rather than silently dropped, so a
corrupt transcript folder is diagnosable instead of just quietly smaller.

## Notes

- Pre-existing on `main`; unrelated to #423, though found while building it.
- Symbols rather than line numbers throughout, since the surrounding code is
  actively changing on the #423 branch.

edspencer:
Fixes #424.

## Problem

A single unreadable entry inside a Claude Code transcript folder — the simplest repro being a **directory** named `<uuid>.jsonl` — made session listing throw outright:

- **`getAgentSessions`** lost the agent's whole list (the throw rejected `mapWithConcurrency`, which rejects like `Promise.all`).
- **`getAllSessions`** lost **every** agent's list, because the throw escaped a per-session loop that is itself nested in the loop over directories.

Callers got an exception, not a partial result — good transcripts sitting next to the bad entry were lost.

### Why the existing guards miss it

On Linux `open(2)` on a directory **succeeds** and only `read(2)` fails, so `createLineReader`'s `stream.on("error")` handler never fires — the `EISDIR` surfaces later on the `for await` inside `isSidechainSession`. `listSessionFiles` can't screen it either: it filters on the `.jsonl` **name**, and `stat()` on a directory succeeds, so the entry is admitted with a valid mtime and size.

## Fix

The principle from the issue: **enrichment of ONE session must never take down the listing.** Both listing paths now isolate per-entry enrichment in a `try/catch`, skip the bad entry, and `logger.warn` it. This covers **all three** enrichment doors (`resolveSidechain`, `resolveAutoName`, `resolvePreview`), not just the sidechain check that happens to throw first — wrapping the whole per-session body closes every door at once.

- `getAgentSessions`: per-session work extracted into `enrichAgentSession`; the `mapWithConcurrency` callback wraps it and returns a skip sentinel (`{ session: null }`, no cache updates) on failure.
- `getAllSessions`: the per-session loop body is wrapped; a failure logs and continues to the next entry (and next directory).

The `sessionBelongsToWorkingDirectory` / `readSessionCwd` disambiguation path was already hardened (it uses `Promise.race` + `try/catch`), so it needed no change.

## Design decision — report vs. silently drop

The issue asked me to decide deliberately whether a skipped entry should be *reported* via a `skipped`/`errors` channel on the result, or dropped.

**Chosen: skip + `logger.warn`, no change to the public return shape.** Reasoning:

1. `getAgentSessions` returns `DiscoveredSession[]` and `getAllSessions` returns `DirectoryGroup[]` — both public and consumed by the thin CLI/web clients. Adding a channel means changing those return types, which is a **breaking change requiring a major bump** for what is a bug fix (repo follows semver strictly).
2. A `warn` log carrying the entry path + reason makes a corrupt folder **fully diagnosable** — meeting the "diagnosable instead of quietly smaller" goal — without the API churn.
3. It matches this file's existing convention: `listSessionFiles` / `getAllSessions` already `logger.warn`/`logger.debug` on `readdir`/`stat` failures rather than surfacing them in the return value.

If a structured channel is wanted later, it can land as its own deliberate major.

## Tests

New `session-discovery-unreadable.test.ts` reproduces the **genuine** `EISDIR` (real `jsonl-parser`, a poison directory next to a good transcript) for both entry points, plus a multi-directory case proving one bad entry no longer aborts sibling directories. Confirmed **red before / green after** — all three fail with `Error: EISDIR: illegal operation on a directory, read` on `main`, and pass with the fix.

> Note: the test mocks `cli-session-path`'s `getCliSessionDir`/`getCliSessionFile` to route through the temp `claudeHomePath`. Those helpers derive their path from `os.homedir()`, not the service's `claudeHomePath`; in production the two coincide (so the poison dir found by the listing is the same file enrichment reads), and the mock restores that invariant under test — otherwise enrichment reads the real home and never hits the fixture.

## Verification

- `pnpm --filter @herdctl/core test` — 3484 passed; the **only** failure is the pre-existing `directory.test.ts > "throws StateDirectoryCreateError when parent directory is not writable"` (runs-as-root chmod artifact, unrelated).
- `biome check` on the changed files — clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Session listings now remain available when individual transcript entries are unreadable.
  * Invalid or inaccessible entries are skipped with a warning instead of preventing other valid sessions from appearing.
  * Session discovery continues across project directories, including when some entries cannot be processed.

* **Tests**
  * Added coverage for unreadable transcript files and directories masquerading as transcript files.

### DIFF
--- a/.changeset/beige-lizards-flow.md
+++ b/.changeset/beige-lizards-flow.md
@@ -1,0 +1,12 @@
+---
+"@herdctl/core": patch
+---
+
+Harden session discovery against a single unreadable transcript entry. An entry that
+`stat()`s as a valid `.jsonl` file but is actually a directory (so `open(2)` succeeds
+and `read(2)` throws `EISDIR`) previously threw out of the per-session enrichment and
+took down the whole listing — `getAgentSessions` lost the agent's entire list, and
+`getAllSessions` lost every agent's list because the throw escaped a loop nested in the
+loop over directories. Both paths now isolate per-entry enrichment (sidechain / auto-name
+/ preview) in a try/catch, skip the bad entry, and log a warning so a corrupt transcript
+folder is diagnosable, while good transcripts next to it still come back.
--- a/packages/core/src/state/__tests__/session-discovery-unreadable.test.ts
+++ b/packages/core/src/state/__tests__/session-discovery-unreadable.test.ts
@@ -1,0 +1,214 @@
+import { mkdir, realpath, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// =============================================================================
+// Mocks
+//
+// This suite deliberately leaves the REAL `jsonl-parser` in place so the
+// enrichment path streams real files. That's the whole point: a directory named
+// `<uuid>.jsonl` reproduces the genuine EISDIR that `open(2)`-succeeds /
+// `read(2)`-fails yields on Linux (issue #424) — a mock can't reproduce that.
+//
+// `cli-session-path` is mocked in ONE respect only: `getCliSessionDir` /
+// `getCliSessionFile` normally derive their path from `os.homedir()`, NOT from
+// the service's `claudeHomePath`. In production those coincide (`claudeHomePath`
+// IS `~/.claude`), so the poison directory that the listing finds is the same
+// file the enrichment then reads. Under test we point both at the temp fixture
+// via `hoisted.claudeHome`; everything else in the module stays real.
+//
+// Attribution and the metadata store ARE mocked: attribution so the good
+// transcript is attributed to the agent under test, and the metadata store so
+// every cache lookup misses (forcing real extraction) without touching disk.
+// =============================================================================
+
+const hoisted = vi.hoisted(() => ({ claudeHome: "" }));
+
+vi.mock("../../runner/runtime/cli-session-path.js", async (importActual) => {
+  const actual = await importActual<typeof import("../../runner/runtime/cli-session-path.js")>();
+  const { join: joinPath } = await import("node:path");
+  return {
+    ...actual,
+    getCliSessionDir: (workspacePath: string) =>
+      joinPath(hoisted.claudeHome, "projects", actual.encodePathForCli(workspacePath)),
+    getCliSessionFile: (workspacePath: string, sessionId: string) =>
+      joinPath(
+        hoisted.claudeHome,
+        "projects",
+        actual.encodePathForCli(workspacePath),
+        `${sessionId}.jsonl`,
+      ),
+  };
+});
+
+vi.mock("../session-attribution.js", () => {
+  const fn = vi.fn();
+  return {
+    buildAttributionIndex: fn,
+    AttributionIndexBuilder: class MockAttributionIndexBuilder {
+      build = fn;
+    },
+  };
+});
+
+const mockGetCustomName = vi.fn().mockResolvedValue(undefined);
+const mockGetAutoName = vi.fn().mockResolvedValue(undefined);
+const mockBatchSetAutoNames = vi.fn().mockResolvedValue(undefined);
+const mockGetPreview = vi.fn().mockResolvedValue(undefined);
+const mockBatchSetPreviews = vi.fn().mockResolvedValue(undefined);
+const mockGetSidechain = vi.fn().mockResolvedValue(undefined);
+const mockBatchSetSidechains = vi.fn().mockResolvedValue(undefined);
+const mockGetUsage = vi.fn().mockResolvedValue(undefined);
+const mockSetUsage = vi.fn().mockResolvedValue(undefined);
+vi.mock("../session-metadata.js", () => ({
+  SessionMetadataStore: class MockSessionMetadataStore {
+    getCustomName = mockGetCustomName;
+    getAutoName = mockGetAutoName;
+    batchSetAutoNames = mockBatchSetAutoNames;
+    getPreview = mockGetPreview;
+    batchSetPreviews = mockBatchSetPreviews;
+    getSidechain = mockGetSidechain;
+    batchSetSidechains = mockBatchSetSidechains;
+    getUsage = mockGetUsage;
+    setUsage = mockSetUsage;
+  },
+}));
+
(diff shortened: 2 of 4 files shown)


edspencer (commented):
**Hardening session discovery against a single unreadable transcript entry (issue #424).** The fix isolates per-entry enrichment in try/catch on both listing paths so one bad entry (e.g. a directory named `<uuid>.jsonl` that `stat()`s as a file but throws `EISDIR` on read) is skipped + logged rather than aborting the whole listing.

**What I checked:**
- Read the full changed region of `session-discovery.ts` (both `getAgentSessions`/`enrichAgentSession` at ~672–834 and the `getAllSessions` enrichment loop at ~1060–1201).
- Confirmed the extraction is behavior-preserving: the new `enrichAgentSession` body is identical to the removed inline block, and the reducer at 707–722 handles the `{ session: null }` skip sentinel (drops the entry, applies no cache updates).
- Verified `continue` statements remain valid inside the new try block, and the `EISDIR` surfaces at the first read door (`resolveSidechain`, line 1074) before `visibleSessionCount++`, so the skipped poison entry isn't counted.
- Confirmed `logger` (line 130) and the `AttributionIndex` type (line 32) both resolve.
- Reviewed the new test suite: it deliberately keeps the real `jsonl-parser` and reproduces the genuine `EISDIR` with a poison directory next to a good transcript, covering both entry points plus the multi-directory sibling case.

Design tradeoff (documented in the PR): the catch-all also swallows transient read errors (EMFILE/ENOSPC), which could intermittently drop otherwise-valid sessions — but each is `logger.warn`'d so a corrupt/degraded folder stays diagnosable, and this matches the file's existing convention. Reasonable for a patch that must avoid the breaking API change a structured errors channel would require. No blocking issues found.

edspencer:
Reviewed 2 changed files (4 hunks); 0 findings.

edspencer:
Addressed the `visibleSessionCount` over-count in 51469d4: an entry counted as visible before a later whole-file read fails (auto-name/preview) is now backed out in the catch, so `sessionCount` stays in sync with the returned sessions. Added a regression test for the deeper-read-failure path (red-before/green-after). Skipped the ast-grep "sensitive data in log" note — the warn logs a transcript path + error message, consistent with the surrounding `readdir`/`stat` warnings in this file.

edspencer (commented):
**Follow-up to the #424 hardening: keep `sessionCount` in sync when an entry is skipped after being counted.** In `getAllSessions`, a transcript can pass the first-line sidechain check (incrementing `visibleSessionCount`) yet later throw on a whole-file read for auto-name/preview, landing in the catch that skips it. The PR adds a `countedAsVisible` flag, set right after `visibleSessionCount++`, and decrements the count in the catch so `sessionCount` matches the sessions actually returned.

**What I checked:**
- Read the full enrichment loop (`session-discovery.ts` 1060–1200). Confirmed `countedAsVisible` is set `true` only immediately after `visibleSessionCount++` (line 1120) and is the sole gate on the decrement (1186–1188).
- Verified the pairing is exact: `sessions.push` is the last statement in the try (no throw can occur between push and success), and every non-error early-exit (`isSidechain`, disambiguation `!belongs`, docker attribution mismatch, pagination `selectedSessionIds` miss) either happens before the increment or is a plain `continue` that never reaches the catch — so no double-decrement and no stranded count.
- Reviewed the added test: both sessions pass the mocked sidechain check, and `resolveAutoName` rejects only for `session-bad`, asserting `sessionCount === sessions.length === 1`. It targets precisely the "counted-then-failed" path.

Small, correct, well-tested. No blocking issues; nothing to flag even at low severity. (Note: the count is best-effort under simultaneous pagination + corruption — a corrupt entry outside the paginated window is never read so stays counted — but that's an acceptable pre-existing edge, not introduced here.)

Issue #427, opened 2026-08-01
perf(core): setUsage has no batch variant — warming one agent's sessions rewrites the whole metadata file N times (~493 MB, ~9.6 s)


edspencer:
`autoName`, `preview` and `isSidechain` all have `batchSet*` methods on `SessionMetadataStore`. **`usage` does not** — `batchSetUsage` does not exist.

Every `setUsage` call therefore runs the full cycle:

```ts
async setUsage(agentName, sessionId, usage, mtime) {
  let metadata = await this.loadMetadata(agentName);      // read whole file
  if (!metadata) metadata = this.createEmptyMetadata(agentName);
  metadata.sessions[sessionId] = { ...sessionEntry, usage, usageMtime: mtime };
  await this.saveMetadata(agentName, metadata);           // atomicWriteJson: rewrite WHOLE file
}
```

And it is called **once per session**, from `getSessionUsage` (`session-discovery.ts:951`), immediately after each extraction — reached per-session via `FleetManager`. So warming N cold sessions performs N whole-file reads and N whole-file writes: **write volume is O(N x filesize), i.e. quadratic in session count.**

## Measured

Against a copy of a real `session-metadata/<agent>.json` from a production instance (**293 KB, 1,677 sessions**), simulating load -> mutate -> atomic write:

| | |
|---|---|
| per `setUsage` | **5.71 ms** |
| to warm all 1,677 sessions | **~9.6 s, ~493 MB written** |

An independent measurement during the same review, on the same file at an earlier size (261 KB / 1,520 sessions), got **2.99 ms per write / 389 MB / ~4.5 s**. The two differ by roughly 2x on per-write cost — almost certainly methodology (whether the read is counted, and `JSON.stringify` indentation) — but agree on the shape and the order of magnitude. Either way it is hundreds of MB of writes to populate a cache that holds a few hundred KB.

Reads are **not** the problem: a full `JSON.parse` + Zod `safeParse` of that file is ~3 ms, and ~20 ms even at 3.5 MB. This is purely write amplification.

## Why it matters now

1. **It is already happening in production** on every cold start that enumerates sessions.
[...]

Issue #426, opened 2026-08-01
bug(core): usage cache validates forever when mtime moves backwards — '>=' should be '===', and the key needs size


edspencer:
`packages/core/src/state/session-discovery.ts` (`getSessionUsage`, ~line 947 in the built output):

```ts
if (cached?.usage && cached.usageMtime && cached.usageMtime >= mtimeStr) {
  return cached.usage;
}
const usage = await extractSessionUsage(filePath);
await this.sessionMetadataStore.setUsage(agentName, sessionId, usage, mtimeStr);
```

Two defects in the freshness check, in increasing severity.

## 1. `>=` should be `===` — a stale entry can validate permanently

The comparison accepts a cached entry stamped **newer** than the file it describes. Once that happens the entry is pinned forever: no later mtime can ever be greater, so the cache never expires and `extractSessionUsage` is never called again. The wrong token totals are served indefinitely, with **no error and no expiry**.

This is not hypothetical — transcript mtimes move backwards routinely:

- `cp` or `rsync` without `-p`
- a **restic restore** (on the instance I measured, `.chats/` is covered by restic snapshots, and restoring is the documented recovery path)
- copying a data dir to make a clone for testing — the perf clone on that box was made exactly this way

Any of those produce a file older than its cache entry, and from then on the usage numbers are silently frozen at pre-restore values. A restore is precisely the moment you most need correct state.

`===` is the correct predicate: the cache describes *this exact version* of the file, not 'some version no newer than this'.

## 2. The key should be `{mtime, size}`, not mtime alone

mtime alone cannot distinguish a file that changed within the filesystem's timestamp granularity, and it is inconsistent with the rest of the codebase — core's own `job-index.ts` keys on mtime **and** size, and paddock's `JobsDirIndex` does too. Only `session-metadata` and `AttributionIndexBuilder` rely on mtime alone.

Adding `size` is also a **prerequisite for resumable extraction**, which is the direction this cache is heading (transcripts are append-only, so a cache entry of `{mtime, size, byteOffset, runningTotals}` lets a grown transcript be parsed from `byteOffset` instead of re-read in full: `size > byteOffset` → parse the delta, `size < byteOffset` → full recompute). Size is load-bearing there regardless, so it is worth adding now rather than changing the key twice.

## Ask

- Change `>=` to `===`.
- Extend the cache key to `{mtime, size}`; `SessionMetadataEntry` needs a `usageSize` (or an equivalent composite stamp) alongside `usageMtime`.
[...]

PR #429, opened 2026-08-01
fix(core): stop SessionMetadataStore clobbering an unreadable metadata file (#419)

### LINKED ISSUE #419: bug(core): SessionMetadataStore replaces the whole metadata file when a READ fails — an empty or unreadable file silently destroys every customName
## Summary

When `SessionMetadataStore` fails to **read** an agent's metadata file, it treats that as "no metadata exists" and the next write **replaces the whole file**. `customName` is user-authored and not derivable from anything, so a transient read failure silently and permanently destroys every chat name the user set for that agent.

The trigger does not require a schema change or file corruption. **An empty file is enough.**

## The path

`state/utils/reads.ts:439` — an empty file is a *successful* read of `null`:

```ts
// Handle empty file
if (content.trim() === "") {
  return { success: true, data: null as T };
}
```

`state/session-metadata.ts:130-162` — `loadMetadata` returns `null` for a failed read **and** for a schema mismatch, indistinguishably from "file absent":

```ts
if (!result.success) {
  if (result.error.code === "ENOENT") {
    return null;                                   // legitimately absent
  }
  logger.warn(`Failed to read metadata file for ${agentName}: ...`);
  return null;                                     // read FAILED — same answer
}

const parseResult = SessionMetadataFileSchema.safeParse(result.data);
if (!parseResult.success) {
  logger.warn(`Corrupted metadata file for ${agentName}: ... Returning null.`);
  return null;                                     // invalid — same answer
}
```

Every mutating setter then does the same thing (`:219-223`, `:334-337`, `:375-378`, `:441-444`, `:480-483`, `:548-551`, `:660-663`):

```ts
let metadata = await this.loadMetadata(agentName);
if (!metadata) {
  metadata = this.createEmptyMetadata(agentName);
}
...
await this.saveMetadata(agentName, metadata);      // atomicWriteJson — full replace
```

`saveMetadata` → `atomicWriteJson` (`utils/atomic.ts:181`) is a wholesale replacement, not a merge. So the file goes from "unreadable but intact on disk" to "`{version:1, agentName, sessions:{}}`".

## Reachable triggers

1. **Empty / zero-length file** — `safeReadJson` returns `{success:true, data:null}`, `safeParse(null)` fails, wipe. No corruption of content required.
2. **`EACCES` / `EPERM`** — classified non-transient at `reads.ts:449`, returns `success:false`, wipe on next write (where the write itself succeeds).
3. **JSON syntax error** after retries are exhausted.
4. **A future `version: 2`.** `SessionMetadataFileSchema` pins `version: z.literal(1)` (`session-metadata.ts:73`), so the very first schema evolution walks into this. This is what makes it worth fixing *now* rather than when it bites.

Severity is high (unrecoverable loss of user-authored data), probability is currently low (`atomicWriteJson` makes torn writes unlikely and there is no version 2 yet). It is a landmine under the next schema change.

## Proposal

Make `loadMetadata` distinguish **absent** from **unreadable**, and refuse to destroy data it could not read:

- Return a discriminated result (`{ kind: "absent" }` / `{ kind: "loaded", file }` / `{ kind: "unreadable", error }`) rather than `null` for all three.
- On `unreadable`, **fail the write** rather than replacing the file. A failed `setCustomName` is recoverable; a wiped file is not.
- If a future version bump makes an old file genuinely unusable, migrate or side-car it (`<agent>.json.v1.bak`) — never overwrite in place.
- Consider preserving `customName` specifically even on a best-effort partial parse, since it is the only non-derivable field in the schema.

## Also worth fixing while in here

`SessionMetadataStore.cache` (`session-metadata.ts:117`, read at `:132`) has **no TTL, no file-mtime check, and no public clear**. Once an agent's file is loaded it is never re-read from disk for the process's lifetime, so an out-of-process write is invisible forever. `FleetManager` works around this by sharing a single store instance (`fleet-manager.ts:1259`), which makes correctness depend on every writer going through the same object.

---

*Found while auditing the cache/index layer for a Paddock performance workstream. No exploit or data loss observed in production — this is a code-reading finding with a verified reproduction path.*

### LINKED ISSUE #424: One unreadable entry in a transcript folder throws and blanks the entire session listing

edspencer:
Fixes #419.

## The bug (verified against `src/`, not `dist/`)

`SessionMetadataStore.loadMetadata()` collapsed **three distinct outcomes into the same `null`**:

1. **ENOENT** — file absent. Legitimate; storage is sparse by design.
2. **Any other read error** (EACCES, EIO, EISDIR, a truncated read that never settled) — logged `warn`, returned `null`.
3. **Schema validation failure** (`safeParse`) — logged, returned `null`.

Every one of the **7 setters** then did `metadata ??= createEmptyMetadata()` and called `saveMetadata` → `atomicWriteJson`, which **replaces the whole file**. So for cases 2 and 3, the next write silently replaced the entire agent file with a fresh empty one, destroying every `customName`, `preview`, `autoName`, `isSidechain` and `usage` entry for that agent. Because the atomic write *succeeds*, it looked like a clean write — no error surfaced.

A single failed read followed by one write is enough, and **a session listing warming its caches performs exactly that read-then-write** — so this triggers in normal operation, not just under exotic disk faults.

Ed's report was accurate. One correction: the in-memory `this.cache` is typed `Map<string, SessionMetadataFile>` and **only ever stored successfully-parsed files** — it never cached `null`, so a transient failure wasn't already sticky. The fix preserves that property (and a test pins it down).

## The fix

- Reads and writes now go through an internal discriminated `readMetadata()` returning `{ status: "loaded" | "absent" | "unreadable" }`.
- **Absent → `createEmptyMetadata()`** — unchanged, correct for sparse storage.
- **Unreadable/corrupt → refuse to write.** A new `loadForWrite()` (used by all 7 setters) throws `SessionMetadataUnreadableError` **without writing**, leaving the damaged file byte-for-byte on disk so it stays recoverable.
- **Getters unchanged.** `loadMetadata()` still returns `null` on absent *or* unreadable, so `getCustomName`/`getAgentMetadata`/etc. degrade gracefully instead of throwing.
- **`session-discovery` background cache-warming** (`batchSet*` / `setUsage`) wraps its writes in `persistCacheUpdates`, which swallows only `SessionMetadataUnreadableError` (logs a warn) and rethrows anything else. A poisoned metadata file for one agent no longer aborts a whole listing — the cache just stays cold and re-extracts next time. This matters because discovery's cache-warming is itself a primary trigger of the bug.
- **User-initiated writes** (`fleet-manager.setSessionName` → `setCustomName`) let the error propagate, so a failure to persist a name surfaces loudly rather than nuking the file.
- `SessionMetadataUnreadableError` and `isSessionMetadataUnreadableError` are exported from `@herdctl/core`.

## Tests

New `#419` describe block in `session-metadata.test.ts`, each asserting **the file is NOT replaced**:
- absent file → empty metadata created (existing behaviour preserved)
- malformed / truncated JSON → throws, bytes untouched (incl. the discovery `batchSet*` path)
- valid JSON failing `safeParse` (`version: 2`; malformed session entry) → throws, bytes untouched
- non-ENOENT read error (EISDIR via a directory at the file path) → throws with `reason: "read-error"`, obstruction untouched
- **round-trip:** a cold-cache store attempting a write against a truncated-but-recoverable file preserves it byte-for-byte
- getters still return `undefined`/`null` (no throw) for an unreadable file
- a transient failure is not sticky — the next call recovers once the file is readable again

`pnpm typecheck`, `pnpm build`, and the core suite pass (3494 passed). The one failing test locally — `directory.test.ts › throws StateDirectoryCreateError when parent directory is not writable` — is pre-existing and environmental: it `chmod 0o444`s a dir and expects a write to fail, which root bypasses. It's untouched by this change and passes in CI (non-root).

## Relationship to #428 (`fix/424-unreadable-session-entry`)

Overlapping file, independent bug. **#428 fixes #424** ("one unreadable transcript entry throws and blanks the whole listing") by isolating per-entry *read* enrichment in try/catch. **This PR fixes #419** ("a failed *read* makes the metadata file get overwritten with an empty one") in `session-metadata.ts`, which #428 does not touch (its test mocks the store out entirely, so it never exercises the overwrite path — I verified #428 does **not** fix #419).

The two are complementary and follow the same house convention #428 settled on — *skip/tolerate + `logger.warn`, don't blank a listing* — for the background path; the only place this PR throws is the store setter, so genuine user writes fail loudly instead of destroying data. Both edit `session-discovery.ts`, so whichever lands second will need a trivial merge resolution (my `persistCacheUpdates` wraps the batch writes that run *after* #428's per-entry try/catch). Happy to rebase on #428 if it lands first.

## Deliberately NOT bundled (separate tickets)

- **#426** — `session-discovery.ts` `usageMtime >= mtimeStr` should be `===` (and keyed on `{mtime, size}` like `job-index.ts`); `>=` validates an entry stamped *newer* than its file forever, and mtime moves backwards on `cp`/`rsync` without `-p` and on restic restores. Lives in the same file #428 edits.
- **#427** — `setUsage` has no `batchSetUsage`, so every call rewrites the whole file (~389 MB of writes to warm 1,520 sessions).

Both are genuinely separable and neither blocks #419 landing on its own, so they're left out here.

### DIFF
--- a/.changeset/fix-419-metadata-no-clobber.md
+++ b/.changeset/fix-419-metadata-no-clobber.md
@@ -1,0 +1,32 @@
+---
+"@herdctl/core": patch
+---
+
+fix(core): stop SessionMetadataStore from replacing an unreadable metadata file with an empty one (#419)
+
+`loadMetadata()` collapsed three outcomes into the same `null`: the file was
+absent (legitimate — storage is sparse), the file could not be read (EACCES /
+EIO / EISDIR / a truncated read), or the file parsed but failed schema
+validation. Every setter treated `null` as "start fresh", so the next
+`atomicWriteJson` replaced the **whole** agent file with an empty structure —
+silently destroying every `customName`, `preview`, `autoName`, `isSidechain`
+and `usage` entry for that agent. One failed read followed by one write (a
+listing warming its caches is enough) wiped the file, and because the atomic
+write succeeded it looked clean.
+
+Reads and writes now distinguish absent from unreadable:
+
+- **Absent** → an empty structure is created, exactly as before.
+- **Unreadable / corrupt** → writes refuse and throw the new
+  `SessionMetadataUnreadableError`, leaving the damaged file untouched so its
+  bytes stay recoverable. Getters keep returning `null`/`undefined` (graceful
+  read degradation is unchanged).
+
+Session-discovery's background cache-warming (`batchSet*` / `setUsage`) tolerates
+that refusal and continues — a poisoned metadata file for one agent no longer
+aborts a whole listing; the cache just stays cold until the file is repaired.
+The transient failure is never cached, so a temporary read error does not become
+sticky.
+
+`SessionMetadataUnreadableError` and `isSessionMetadataUnreadableError` are
+exported from `@herdctl/core`.
--- a/packages/core/src/state/__tests__/session-discovery-combined-424-419.test.ts
+++ b/packages/core/src/state/__tests__/session-discovery-combined-424-419.test.ts
@@ -1,0 +1,207 @@
+import { mkdir, readdir, readFile, realpath, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// =============================================================================
+// #424 × #419 — both failure modes live at once
+//
+// This suite proves the two fixes *cooperate* rather than merely coexist. It is
+// deliberately a HIGH-fidelity integration test:
+//
+//   - The REAL `jsonl-parser` is used (no mock) so a directory named
+//     `<uuid>.jsonl` reproduces the genuine `open(2)`-succeeds / `read(2)`-throws
+//     EISDIR that only surfaces at read time on Linux — the #424 trigger.
+//   - The REAL `SessionMetadataStore` is used (no mock) so a corrupt
+//     `session-metadata/<agent>.json` reproduces the genuine #419 trigger: the
+//     store must refuse to overwrite it when the listing warms its caches.
+//
+// Only `session-attribution` and `cli-session-path` are stubbed, exactly as the
+// #424 suite (`session-discovery-unreadable.test.ts`) does: attribution so the
+// good transcript is attributed to the agent under test, and `getCliSessionFile`
+// / `getCliSessionDir` so per-session reads resolve to the temp fixture instead
+// of the real `os.homedir()`. In production those coincide; the mock restores
+// that invariant AND keeps the test off the live `~/.claude` tree.
+// =============================================================================
+
+const hoisted = vi.hoisted(() => ({ claudeHome: "" }));
+
+vi.mock("../../runner/runtime/cli-session-path.js", async (importActual) => {
+  const actual = await importActual<typeof import("../../runner/runtime/cli-session-path.js")>();
+  const { join: joinPath } = await import("node:path");
+  return {
+    ...actual,
+    getCliSessionDir: (workspacePath: string) =>
+      joinPath(hoisted.claudeHome, "projects", actual.encodePathForCli(workspacePath)),
+    getCliSessionFile: (workspacePath: string, sessionId: string) =>
+      joinPath(
+        hoisted.claudeHome,
+        "projects",
+        actual.encodePathForCli(workspacePath),
+        `${sessionId}.jsonl`,
+      ),
+  };
+});
+
+vi.mock("../session-attribution.js", () => {
+  const fn = vi.fn();
+  return {
+    buildAttributionIndex: fn,
+    AttributionIndexBuilder: class MockAttributionIndexBuilder {
+      build = fn;
+    },
+  };
+});
+
+import { buildAttributionIndex } from "../session-attribution.js";
+import { SessionDiscoveryService } from "../session-discovery.js";
+
+const mockBuildAttributionIndex = vi.mocked(buildAttributionIndex);
+
+// =============================================================================
+// Helpers
+// =============================================================================
+
+async function createTempDir(prefix: string): Promise<string> {
+  const baseDir = join(tmpdir(), `${prefix}-${Date.now()}-${Math.random().toString(36).slice(2)}`);
+  await mkdir(baseDir, { recursive: true });
+  return await realpath(baseDir);
+}
+
+/** A minimal but valid transcript: one plain user line, not a sidechain. */
+async function writeGoodTranscript(dir: string, sessionId: string, cwd: string): Promise<void> {
+  const line = JSON.stringify({
+    type: "user",
+    isSidechain: false,
+    timestamp: "2026-08-01T00:00:00.000Z",
+    cwd,
+    message: { role: "user", content: "hello from a healthy transcript" },
+  });
+  await writeFile(join(dir, `${sessionId}.jsonl`), `${line}\n`);
+}
+
+/** A poison entry: a DIRECTORY named `<uuid>.jsonl` — the #424 trigger. */
+async function writePoisonEntry(dir: string, sessionId: string): Promise<void> {
+  await mkdir(join(dir, `${sessionId}.jsonl`), { recursive: true });
+}
+
+function mockAttribution(attributedIds: Set<string>, agentName: string) {
+  const getAttribute = (sessionId: string) => ({
+    origin: "native" as const,
+    agentName: attributedIds.has(sessionId) ? agentName : undefined,
+    triggerType: undefined,
+  });
+  return {
+    getAttribute,
+    getAttributes: (ids: string[]) => new Map(ids.map((id) => [id, getAttribute(id)])),
+    size: 0,
+  };
+}
+
+// =============================================================================
+// Tests
+// =============================================================================
+
+describe("SessionDiscoveryService — #424 poison entry AND #419 corrupt metadata, simultaneously", () => {
+  let tempHome: string;
+  let claudeHomePath: string;
+  let tempStateDir: string;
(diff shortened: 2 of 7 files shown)


edspencer (commented):
Solid, well-tested fix for #419. The change replaces the old three-way-collapsed `loadMetadata() → null` with an internal discriminated `readMetadata()`, adds `loadForWrite()` that refuses (throws `SessionMetadataUnreadableError`) rather than clobbering an unreadable/corrupt file, keeps getters graceful, and has `session-discovery` swallow only that error during cache-warming. I verified: the seven setters now route through `loadForWrite`; the `reason` classification matches `SafeReadError.code` (I read `reads.ts` — parse errors carry no `code` → `corrupt`, I/O errors carry an errno → `read-error`); the discovery `persistCacheUpdates` wrapper covers all `batchSet*` and `setUsage` call sites; and `fleet-manager.setSessionName → setCustomName` intentionally lets the error propagate. One asymmetry noted below.

edspencer:
Checked the test against the real source: `SessionDiscoveryService` constructor accepts `{ claudeHomePath?, stateDir }` and instantiates a real `SessionMetadataStore(stateDir)` when none is injected (session-discovery.ts:240-245), so the store is genuinely unmocked. `listSessionFiles` filters by `.jsonl` name suffix then `stat`s (session-discovery.ts:354-363), so a directory named `<uuid>.jsonl` is enrolled as a session and only fails at read time (EISDIR) — the real #424 trigger, not a filtered-out no-op. `metadataKey` resolves to `agent.name` for the getAllSessions fleet-directory path (line 1053) and the raw `agentName` for getAgentSessions, both matching the seeded `session-metadata/my-agent.json`. Warming the good session's caches produces a needed autoName/sidechain update, so `persistCacheUpdates` → `batchSet*` → `loadForWrite` reads the corrupt file and throws `SessionMetadataUnreadableError` before `atomicWriteJson` runs — hence the file stays byte-for-byte intact and no temp file is created, making all three assertions (survival, sessionCount sync, single-file-remaining) meaningful. No correctness, flakiness, or isolation problems observed (each test builds a fresh service + temp dirs).

Reviewed 1 changed file (1 hunk); 0 findings.

edspencer (commented):
Defensive refactor of session discovery (issue #424, composing with #419). `getAgentSessions` extracts the per-session enrichment into `enrichAgentSession` and wraps each call in a `try/catch` inside `mapWithConcurrency`; `getAllSessions` wraps its inner per-entry loop body in a `try/catch` with a `countedAsVisible` flag that backs out `visibleSessionCount` on failure. Both skip the bad entry and `logger.warn`, so an entry that `stat()`s as `.jsonl` but is a directory (EISDIR on read) no longer takes down the whole listing. No substantive issues found.

edspencer (commented):
Test-only PR: adds a high-fidelity integration test (`session-discovery-combined-424-419.test.ts`) that exercises the #424 poison-entry skip and the #419 refuse-to-clobber paths simultaneously, using the real `SessionMetadataStore` and real `jsonl-parser` (only attribution and `cli-session-path` are stubbed). I verified the test genuinely drives the code it claims to and found no issues.

Issue #357, opened 2026-07-11
CLI runtime: resume:null session-id resolution steals a co-located agent's session id (shared cwd + concurrent stream)


edspencer:
## Summary

`waitForNewSessionFile()` (`packages/core/src/runner/runtime/cli-session-path.ts`, compiled `dist/runner/runtime/cli-session-path.js:336`) resolves the session id of a freshly-spawned `resume:null` CLI turn by picking the **newest `.jsonl` whose `mtime > startTime`** in the agent's session directory. When two agents **share a working directory** (hence the same `~/.claude/projects/<encoded-cwd>/` session dir), a concurrently-**streaming** session from the *other* agent has `mtime > startTime` too — and can be newer than the file the CLI just created. The new turn is then mis-resolved to the **other agent's** session id, and the job record is written with that foreign `session_id`.

## Impact

The job/attribution store gets a record `{ session_id: <other agent's session>, agent: <this agent> }`. Because `buildJobIndex` (`state/session-attribution.js`) is **last-writer-wins per `session_id`**, `getAttribute(session)` then resolves the victim session to the wrong agent, and `SessionDiscoveryService.getAgentSessions(victimAgent, …)` filters it out (`session-discovery.js` — `if (attribution.agentName !== agentName) continue;`). The victim's chat disappears from its owner's session list until a later turn on that session re-writes a correctly-attributed job (so it's **intermittent**).

## Real-world reproduction (Paddock)

Paddock runs a per-project **keeper** agent and a per-project **sweeper** (post-turn curator) that **share the project directory as cwd**. Observed on a live instance:

- Keeper chat `00d3b8ef-…` (actively streaming).
- A sweep runs (`resume:null`, `trigger_type: manual`) at 21:34–21:35 while the keeper chat streams.
- The sweep's job record: `agent: sweeper-paddock`, `session_id: 00d3b8ef-…` — i.e. it grabbed the **keeper's** active session id (the sweeper's own stream reported no session id; the id came purely from the newest-file heuristic).
- Result: the keeper chat flips to `sweeper-paddock` in the attribution index and vanishes from the keeper's chat list, reappearing after the next keeper turn.
- Across the instance, **7 sessions** currently carry both a keeper job and a colliding sweeper job.

## Minimal repro

1. Agent A and agent B configured with the **same** `working_directory`.
2. Start a long streaming turn on A (keep appending to A's transcript).
3. While A streams, run a `resume:null` turn on B.
[...]

Issue #423, opened 2026-07-31
Adopt pre-existing Claude Code CLI sessions into an agent


edspencer:
## Summary

Add a first-class way to **adopt** an existing Claude Code CLI transcript (a
session a user already ran in a terminal, living under
`~/.claude/projects/<encoded-cwd>/<session-id>.jsonl`) into a herdctl agent, so
it is discoverable, attributed, and resumable exactly like a session the agent
created itself.

Today a transcript is only surfaced under an agent if the attribution index maps
its session id to that agent (built from job records / platform records). A
transcript produced outside herdctl has no such record, so it is invisible to
`getAgentSessions` even when it physically sits in the agent's own working-dir
folder.

## Motivation

Users accumulate substantial CLI history before (or alongside) adopting a
server/agent-hosted workflow. There is currently no supported path to bring that
history "into" an agent. The natural user request is: *"I already have these
chats — show them here and let me resume them."*

## Proposed shape

A small, layered design:

1. **A dedicated adoption store**, separate from job records.
   Adoption is *not* a run, so it should not be expressed as a synthetic job
   record. A purpose-built store (`<stateDir>/adopted-sessions/<session-id>.yaml`
   → `{ agentName, adoptedAt, sourceCwd? }`) reads cleanly as a **third source**
   in the attribution index (alongside jobs and platform records), returning a
   new `origin: "adopted"`.

2. **Public API on the fleet/session layer:**
   - `adoptSession(agentName, sessionId, { sourceCwd? })` — record attribution
     (idempotent).
   - `listAdoptableSessions(agentName, fromWorkingDir?)` — native, non-sidechain,
     not-yet-adopted sessions in a working dir's transcript folder.
   - `adoptSessionsFrom(agentName, { fromWorkingDir?, mode?, dryRun? })` — the
     workhorse: for each adoptable session, place the transcript where discovery
     looks (when the source dir differs from the agent's own), then record
     attribution. Returns `{ adopted, skipped }`.

3. **Copy is the default placement mode** (`copy | move | link`). The user's
   original `~/.claude` transcripts must never be mutated by default; the agent
   gets its own copy to append to on resume.

## Gotchas / risks to address

These are non-obvious and easy to get wrong; enumerating them so they are handled
deliberately.

1. **Claude-home resolution is inconsistent across the codebase.**
   Session *discovery* honours an injectable `claudeHomePath`, but several
   sibling helpers resolve the transcript path from the ambient
   `os.homedir()/.claude` instead:
[...]

PR #431, opened 2026-08-01
fix(core): mint the CLI session id instead of inferring it (#357)

edspencer:
Reopens and finishes **#357**. Found via [paddock#548](https://github.com/edspencer/paddock/issues/548).

## What #357 left behind

#357 fixed the case where a co-located agent is **appending** to its own session:
a pre-spawn snapshot identifies the new transcript by set difference instead of
by mtime. It did not fix the case where a co-located agent **creates** a file,
and said so in a comment:

```js
// Normally exactly one; if several appeared (e.g. multiple co-located
// spawns raced), the newest is ours.
const newest = await newestSessionFileByMtime(sessionDir, brandNew);
```

The newest is not ours — it is whoever spawned last. Two agents sharing a
`working_directory` share one `~/.claude/projects/<encoded-cwd>/`, so two
concurrent `resume:null` spawns produce two brand-new files that are
indistinguishable by name *or* mtime. The agents **trade session ids**: each
adopts the other's transcript and writes a job record claiming it.

Reproduced against a build of `main` (5.27.0), no paddock and no vitest —
pre-existing file, snapshot, write `bbbb.jsonl`, sleep 20 ms, write `cccc.jsonl`:

```
returned cccc.jsonl, should be bbbb.jsonl -> WRONG SESSION
```

## The fix: stop inferring

Claude Code accepts `--session-id <uuid>` and names the transcript after it. I
verified both that the file is named exactly as requested and that it composes
with `--fork-session` (the fork wrote the requested id).

So a turn that starts a new session — a fresh run, or a fork — now mints its own
id and passes it. The transcript is known **by name**: no snapshot, no mtime, no
tie-break. A plain resume already knows its id and is untouched.

Two details that matter:

- **While polling, the inference paths are disabled.** Falling back to "take
  whichever brand-new file shows up" mid-wait would reintroduce exactly the
  collision this removes. Only once the deadline passes without our file
  appearing — a CLI too old to know the flag — do we warn and fall back, so the
  turn still runs rather than failing.
- `HERDCTL_CLI_MINT_SESSION_ID=0` restores the old behaviour.

Verified under the hostile ordering: with the co-located agent's file created
**first** *and* newest by mtime, we still take ours.

## Also: #357's "secondary hardening", which was never implemented

#357 proposed it in as many words:

> **Secondary hardening (defense in depth):** in `buildJobIndex`, a job should not
> be allowed to (re)attribute a `session_id` to an agent when the winning record's
> agent differs from an already-known owner…

`AttributionIndexBuilder` filled its session→agent map by iterating a cache
populated inside an unordered `Promise.all`, then `index.set(sessionId, …)` —
last writer wins. So when two records claim one session id, the winner is
**whichever file's stat+parse finished last**.

That is not random, which is worse than it sounds — it is a function of record
size and machine load. With the same two records on disk, enlarging one flips the
result:

| Fixture | Winner |
|---|---|
| keeper record larger (slower parse) | `keeper` 10/10 |
| sweeper record larger | `sweeper` 10/10 |

Claims are now ordered explicitly: **newest `started_at` wins, ties broken on job
id**, so the answer is a pure function of what is on disk.

Newest-wins rather than the literal "first owner keeps it" #357 suggested,
because **adopting or promoting a session is a legitimate change of owner** —
paddock's `writeAgentAdoptionJob` and `reattributeSession` do exactly that, and a
hard owner lock would break them. (`reattributeSession` rewrites every matching
record in place, so a promote leaves no conflict at all; the ordering rule is
what covers adoption and any accidental collision that still gets through.)

## Tests

7 new cases, all of which fail against the current source and pass here:

- the minted id is honoured even when a co-located brand-new file lands first and is newest by mtime
- the deadline fallback still resolves rather than failing the turn
- a contrast case pinning the old inference behaviour (still takes the wrong file)
- attribution resolves identically regardless of the order files are read, is stable across repeated builds, still lets a deliberate re-attribution win, and breaks an exact timestamp tie on job id

`packages/core`: **3488 passed**, 1 pre-existing failure
(`directory.test.ts` "parent directory is not writable" — fails identically on a
clean checkout when the suite runs as root, since root can write to the
restricted dir). Typecheck and lint clean.

## ⚠️ Requires a companion paddock change before paddock upgrades

paddock's integration suite drives a **fake** `claude` (`test/bin/claude`) that
mints its own uuid and ignores `--session-id`. Against this version the expected
file would never appear, so every new-session turn would burn the full 60 s
timeout before falling back — technically passing, practically unusable.

`test/bin/claude` needs to honour `--session-id` when present. Happy to open that
PR alongside; it should land before paddock bumps `@herdctl/core`.

## One thing I could not verify

I don't know which Claude Code version introduced `--session-id` (confirmed
present in 2.1.220), and herdctl declares no minimum. If an older CLI *errors* on
an unknown flag rather than ignoring it, the deadline fallback never gets a
chance and the turn fails loudly. That is the reason for the env escape hatch. If
you know the introducing version, a documented minimum would be better than the
fallback.

## Summary by CodeRabbit

* **New Features**
  * New CLI sessions receive explicit identifiers, improving transcript tracking and attribution.
  * Forked sessions retain reliable links to their source while writing to distinct transcripts.
  * Session discovery gracefully falls back to legacy behavior when explicit identification is unavailable.

* **Bug Fixes**
  * Improved handling when the CLI exits before creating its transcript.
  * Resolved conflicting session claims deterministically, favoring the newest session and providing stable tie-breaking.


packages/core/src/state/session-attribution.ts
```diff
@@ -358,18 +382,42 @@ export class AttributionIndexBuilder {
...
     for (const { entry } of this.jobFileCache.values()) {
-      if (entry) {
-        index.set(entry.sessionId, { agent: entry.agent, triggerType: entry.triggerType });
-      }
+      if (!entry) continue;
+      const held = winners.get(entry.sessionId);
+      if (held && !claimWins(entry, held)) continue;
+      winners.set(entry.sessionId, { startedAt: entry.startedAt, jobId: entry.jobId });
+      index.set(entry.sessionId, { agent: entry.agent, triggerType: entry.triggerType });
     }
```
edspencer, line 550: **LOW · correctness** — Non-incremental buildJobIndex still resolves collisions by order (oldest-wins, non-deterministic on ties)

This PR makes the incremental `AttributionIndexBuilder` resolve a doubly-claimed session id deterministically (newest `started_at` wins, tie-break on job id). But its sibling `buildJobIndex` (lines 100-115), used by the exported `buildAttributionIndex`, was left unchanged: it iterates `result.jobs` and does last-write-wins via `index.set`.

`listJobs` sorts jobs by `started_at` **descending** (job-metadata.ts:443-447), so last-write-wins there resolves a collision to the **oldest** claim — the opposite of the new newest-wins rule this PR argues is correct (adopt/promote should win). And for exact `started_at` ties the result still depends on V8's stable sort preserving `readdir` order, i.e. the same filesystem-order non-determinism this PR set out to remove.

The two builders are documented as parallel (`createAttributionIndex` is "Shared by the full buildAttributionIndex and the incremental AttributionIndexBuilder"), so having them disagree on collision resolution is a latent surprise. `buildAttributionIndex` currently has no production callers besides being a public export, which is why this is low rather than higher — but it's still reachable API. Consider applying the same `claimWins` ordering in `buildJobIndex`, or documenting that the full builder is deprecated in favor of the incremental one.

packages/core/src/runner/runtime/cli-runtime.ts
```diff
@@ -279,6 +280,19 @@ export class CLIRuntime implements RuntimeInterface {
       args.push("--fork-session");
     }
 
+    // A turn that starts a NEW session (a fresh run, or a fork, which writes a
+    // brand-new file) mints its own session id and tells the CLI to use it, so
+    // the transcript is `<mintedId>.jsonl` and we never have to work out which
+    // file was ours (issue #357). A plain resume keeps its existing id and needs
+    // nothing. Opt out with HERDCTL_CLI_MINT_SESSION_ID=0.
+    const mintedSessionId =
+      (!options.resume || options.fork) && process.env.HERDCTL_CLI_MINT_SESSION_ID !== "0"
```
edspencer, line 349: **LOW · bug** — --session-id defaults on; an older CLI that errors on the unknown flag fails every new-session turn

`mintedSessionId` is enabled by default (`HERDCTL_CLI_MINT_SESSION_ID !== "0"`) and pushed as `--session-id`. The deadline fallback in `waitForNewSessionFile` only helps if the CLI *ignores* an unknown flag. If an older `claude` binary instead **errors** on the unknown `--session-id` flag, the subprocess exits non-zero and no fallback runs — every fresh/fork turn fails loudly, by default, with no minimum-CLI-version guard in the repo.

The author acknowledged this in the PR body ("One thing I could not verify"), and the env escape hatch exists — recording it here so it isn't lost in prose. Worth either documenting a minimum Claude Code version, or gating the flag behind a capability probe, before this ships as the default.

edspencer (commented):
This PR fixes co-located session-id collisions by minting the CLI session id (`--session-id <uuid>`) instead of inferring the transcript file, with a deadline fallback to the old inference and an `HERDCTL_CLI_MINT_SESSION_ID=0` escape hatch. It also makes the incremental `AttributionIndexBuilder` resolve doubly-claimed session ids deterministically (newest `started_at` wins, tie-break on job id) instead of by `Promise.all` completion order.

I read `cli-session-path.ts`, `cli-runtime.ts` (spawn/mint/watch flow), `session-attribution.ts`, the `JobMetadataSchema` (confirmed `started_at`/`id` are always present, so `claimWins` never dereferences undefined), and `job-metadata.ts`'s `listJobs` ordering. The mint logic, the "keep waiting for exactly our file while polling" invariant, and the new tie-break all look correct and well-tested. Two lower-severity gaps below.

edspencer:
Verified the three spawn shapes core produces:
- **New** (`--session-id <minted>`, no resume): writes `<minted>.jsonl`, no prior text — this is the case that previously timed out.
- **Resume** (`--resume <id>`, no session-id): `flags.sessionId` is undefined so `??` falls through to `flags.resume`; reads and appends the same file.
- **Fork** (`--resume <src> --fork-session --session-id <new>`): `sessionId` (=`<new>`) wins for the written file, while `priorText` deliberately reads `<src>.jsonl` for continuity.

The key subtlety — writing to the sessionId-named file but reading prior text from the resume source — is handled correctly, and `readPriorText` tolerates a missing source. Transcript `sessionId`/`session_id` fields stay consistent with the filename. Clean pass.

Reviewed 1 changed file (3 hunks); 0 findings.

edspencer:
Pushed `0682236`: the fake `claude` in this repo's own web-UI harness needed the companion change too — I flagged that risk for paddock and then walked straight into it here.

The harness's fake minted its own uuid and ignored `--session-id`, so once the runtime started passing a minted id, every new-session turn waited out the full 60 s timeout before falling back to inference. That is what the `Web UI Tests` failure was: `04-jobs` "triggering a job via the modal" failed outright, `02-schedules` went flaky, and the suite took **8.5 minutes**.

Verified load-bearing rather than assumed: with the fake reverted, that one job test fails and takes 48 s; with the fix, the whole suite is **27 passed in 55 s**. (One flaky — `07-sessions-and-errors:46`, passed on retry. I have not checked whether it is also flaky on `main`, so I am not claiming it is unrelated.)

`--session-id` now names the transcript whenever supplied, winning over `--resume` so a fork lands in a new file rather than appending to its source, with continuity read from the resumed session.

This is the same companion change paddock's `test/bin/claude` will need before it bumps `@herdctl/core`.

edspencer (commented):
Companion test-fixture change: the fake `claude` CLI used by @herdctl/web's UI/integration tests now honours `--session-id`, matching core's new mint-and-pass behaviour (#357). Without it, every new-session turn in the web tests would burn the full 60s timeout waiting for a file named after the minted id the fixture ignored.

I read the full fixture and cross-checked its flag handling against how core actually passes `--resume`/`--fork-session`/`--session-id` (verified in the sibling core PR): new session, plain resume, and fork all resolve to the correct transcript filename and continuity source. No issues found — the change is correct and well-commented.

edspencer:
Rebased onto `main` (was 11 commits behind, at core 5.27.0 → now 5.30.0) and **changed the fallback design**, because the rebase surfaced something worth acting on.

### The conflict was trivial; what it revealed was not

The only conflict was additive-vs-additive in `session-attribution.ts` — #438 added the ownership predicates, this PR added `claimWins`. Both kept, no behaviour reconciled away. Confirmed `main` still has the unfixed assembly loop, so this PR is still needed.

But the rebase also pulled in three test files from #423 (`CLAUDE_CONFIG_DIR` / claude-home threading) whose fake spawners write a session id of their own choosing and ignore `--session-id`. All three failed. Verified the cause exactly: with `HERDCTL_CLI_MINT_SESSION_ID=0`, all 28 of their tests pass.

They were not failing because the fix is wrong. They were failing because **waiting out the full 60 s timeout is an unreasonable cost to impose on every fake-CLI harness** — and this was the third one (herdctl's web-UI fake, these, and paddock's still to come).

### So: wait on evidence, not on the clock

- While the process is **alive**, still refuse to infer. That is the guarantee that matters — a co-located agent's brand-new file must not be claimed just because it landed first.
- Once the CLI has **exited** without writing the file we asked for, it never will. That is settled, not unknown, so fall back immediately (500 ms grace for a late flush).

**All three #423 files now pass unmodified** — no other author's tests touched. Two new tests cover it: the fast fallback completes in <5 s against a 30 s timeout, and the co-located race still resolves to our file while the process is alive.

`packages/core`: **3675 passed**, 1 pre-existing failure (`directory.test.ts` "parent directory is not writable" — fails identically on a clean checkout when run as root). Web UI suite: 27 passed in 59 s. Typecheck and lint clean.

The `test(web)` commit teaching herdctl's own fake to honour the flag is still here and still worth having — the fallback makes a non-honouring fake *survivable*, but honouring it is what makes those runs deterministic.

edspencer:
Brought this branch up to date with current `main` (now includes #446 plugins/MCP passthrough `d45d7f9`, #450, #455, and the releases through #456) so CI validates against the current tree rather than a base predating the plugins/MCP work.

Merged `main` in rather than rebasing — this repo's rules say never force-push, and a rebase of a published branch would require one. Same CI outcome, no rewritten history. Say the word if you'd rather have a clean linear rebase and I'll force-push instead.

Merge was conflict-free. Verified the one flagged interaction: `--plugin-dir` (`cli-runtime.ts:246`) and `--session-id` (`:353`) are independent `args.push` calls with no semantic overlap.

Local gates green before push: `pnpm typecheck` (11/11 tasks) and the two relevant suites — `cli-session-path.test.ts` + `session-attribution.test.ts`, 140 tests passing, including the new `minted-id path (expectedSessionId supplied) — issue #357 follow-up` block.

Also reopened #357 — it was auto-closed by #361, which fixed only the appending-transcript half. The brand-new-transcript race is still live on main (`cli-session-path.ts:442-447`, "the newest is ours"), so merging this will close it against the commit that actually finishes it.

Issue #415, opened 2026-07-28
perf(core): listJobs parses every job record sequentially and filters after parse (~1.2s, 77% discarded)


edspencer:
## Summary

`listJobs()` (`packages/core/src/state/job-metadata.ts`, the `listJobs` loop) reads and
zod-validates **every** `job-*.yaml` in the jobs dir **sequentially**, then applies
`filter.agent` **after** parsing. On a real instance this is ~1.2s of blocking CPU per
call, and 77% of that work is thrown away.

Core already has the right pattern elsewhere — `AttributionIndexBuilder` in
`state/session-attribution.ts` keeps an mtime-keyed `jobFileCache` and only re-parses
files whose `mtimeMs` changed. `listJobs` doesn't use it.

## Measured

Against a production-scale instance (1,996 `job-*.yaml`, 46.6 MB, `@herdctl/core` 5.26.1):

| Call | Time | Records |
|---|---|---|
| `listJobs(dir, { agent: 'keeper-paddock' })` | **1182 ms** | keeps **463 of 1996** parsed → **77% discarded** |
| `listJobs(dir)` unfiltered | **1343 ms** | 1996 |

Three compounding causes:

1. **Sequential** — `for (const file of jobFiles) { await safeReadYaml(...) }`, no concurrency.
2. **Post-filter** — `if (filter.agent && job.agent !== filter.agent) continue;` runs *after*
   `safeReadYaml` + `JobMetadataSchema.safeParse`.
3. **No cache** — every call re-parses from scratch.

The records are also far larger than the fields callers use: on this instance **97.3% of the
bytes are `prompt` (80.6%) + `summary` (16.7%)**, averaging 24.5 KB/record (largest 92 KB),
while list callers typically want `id`/`agent`/`session_id`/`status`/`started_at`/`finished_at`.

## Why it matters downstream

Paddock calls this on hot UI paths — `/api/projects/:slug/runs` (**1.2 s**, fires on every
ProjectView mount, even when the History tab is never opened) and
`/api/projects/:slug/triggers/runtime` (**1.5 s**, re-polled every ~10 s). Because the parse
is synchronous CPU on Node's single thread, these serialize with everything else.

Note the `limit` parameter does not help: it's applied to the *result*, so
`/runs?limit=1` still costs 1.34 s.

## Suggested direction (not prescriptive)

- Reuse the existing incremental mtime-cached index rather than adding a second one.
[...]

PR #432, opened 2026-08-01
feat(state): filter listJobs by a set of agents, and page /api/jobs inside it (#418)

### LINKED ISSUE #418: perf(core): listJobs callers that pass neither filter nor limit hydrate every record — including the web dashboard's paginated /jobs route
## Summary

`listJobs` filters, sorts and pages against the mtime-keyed index added in #415/#416, and hydrates only the records it returns — **but only when a `limit` is passed**. `job-metadata.ts:356`:

```ts
const retain = filter.limit === undefined ? matches : () => false;
```

With no `limit`, `retain = matches`, so every matching record is parsed and held. Several callers pass neither a filter nor a limit and then slice in JS, which fully defeats the index.

## Measured

Against an isolated production-scale copy (2,016 job records, 47.4 MB of `.yaml`), calling the shipped 5.27.0 `listJobs` directly:

| call | warm |
|---|---|
| `listJobs(dir)` — no filter, no limit | **1,106 ms** (cold 1,297 ms — the index buys 15%) |
| `listJobs(dir, { agent })` | 131 ms |
| `listJobs(dir, { limit: 200 })` | **148 ms** |

Heap grows ~150 MB per unfiltered call.

## Callers passing neither filter nor limit

| call site | notes |
|---|---|
| **`packages/web/src/server/routes/jobs.ts:101`** | user-facing paginated endpoint — see below |
| `packages/core/src/fleet-manager/job-manager.ts:551` | `applyRetention` |
| `packages/core/src/fleet-manager/job-manager.ts:585` | `applyRetention` re-fetch |
| `packages/core/src/fleet-manager/log-streaming.ts:241` | when `options.agentName` is undefined |
| `packages/core/src/state/session-attribution.ts:103` | `buildJobIndex`, feeding `buildAttributionIndex` — which has **zero production callers** (see below) |

### The web dashboard's `/jobs` route is the worst of these

`packages/web/src/server/routes/jobs.ts:90-104` builds a filter from `agent`/`status` only, never passes `limit`/`offset`, then paginates in JS:

```ts
const result = await listJobsFn(jobsDir, filter);
const paginatedJobs = result.jobs.slice(clampedOffset, clampedOffset + clampedLimit);
```

So a paginated, user-facing endpoint hydrates the entire jobs directory on every request — exactly the pathology #415 set out to remove, in our own dashboard.

Companion change: `total` is currently `result.jobs.length`; with the push-down it must become `result.total`, which `ListJobsResult` already returns (pre-pagination match count). `packages/web/src/server/__tests__/routes.test.ts:329-332` asserts the current call shape and will need updating.

## Downstream: paddock#535

[edspencer/paddock#535](https://github.com/edspencer/paddock/issues/535) is the same bug in Paddock's `listRunsForAgents`, which needs jobs for a **set** of agents and so passes no filter at all:

```ts
const { jobs } = await listJobs(jobsDir).catch(() => ({ jobs: [], errors: 0 }));
const filtered = jobs.filter((j) => wanted.has(j.agent));
return limit > 0 ? filtered.slice(0, limit) : filtered;
```

Measured downstream effect: `/triggers/runtime` costs **1.27–1.43 s per request and never warms**. It cannot be fixed downstream today because `ListJobsFilter` has no multi-agent filter.

## Proposal

1. **Add `agents?: string[]` to `ListJobsFilter`.** One line in `matches` (`job-metadata.ts:344-350`), one in the interface. Unblocks paddock#535.
2. **Push `limit`/`offset` down in `packages/web/src/server/routes/jobs.ts`**, and switch `total` to `result.total`.
3. Decide on the remaining callers: `applyRetention` genuinely wants everything (fine, but worth a comment saying so); `log-streaming.ts:241` should pass its limit if it has one.
4. **Consider deleting `buildAttributionIndex`** (`session-attribution.ts:191`). It is public API at the package root (`state/index.ts:88`), exercised by ~25 tests and used as a mock anchor in `session-discovery.test.ts`, but has **no production caller** — `SessionDiscoveryService` uses the incremental `AttributionIndexBuilder` instead. Deleting it removes an unfiltered `listJobs` call rather than fixing one. (Breaking change to a public export; may want a deprecation cycle.)

**Note for whoever picks this up:** adding `agents` alone does not fix paddock#535. With `limit` still undefined, `retain = matches` and every match is hydrated. The caller must pass **both** `agents` and `limit`.

## Pre-existing hazard, not introduced by this change

Paddock's `listRunsForAgents` slices to 200 *after* filtering, so a chatty keeper agent can push a rarely-run trigger agent out of the window and the Triggers tab shows it as never-run. Pushing the limit down preserves this behaviour exactly — it neither causes nor fixes it. Worth a separate issue if we want "newest job per agent" as a first-class query.

---

*Measured 2026-07-29 against an isolated read-only copy of a production instance.*

edspencer:
Closes #418 (items 1 and 2). Unblocks edspencer/paddock#535.

## The problem

#415 made `listJobs` index-backed, but only for callers that can express what they want as a **filter**. `ListJobsFilter` had `agent` — exactly one — so a caller needing *several* agents' jobs had no way to say so. It had to list the directory unfiltered and filter in JS:

```ts
const { jobs } = await listJobs(jobsDir);          // hydrates EVERYTHING
const filtered = jobs.filter((j) => wanted.has(j.agent));
return filtered.slice(0, limit);
```

With `limit` undefined, `retain = matches`, so every record in the directory is read, YAML-parsed and Zod-validated — to return a handful.

## The change

**1. `ListJobsFilter` gains `agents?: string[]`.**

- ANDs with `agent` (passing both narrows to the intersection).
- An **empty array matches nothing**, deliberately. The other reading — empty means unfiltered — would silently turn a caller's "I have no agents to look up" back into a full directory scan.

**2. `packages/web`'s `/api/jobs` pages inside `listJobs`** instead of asking for everything and slicing. It is a paginated, user-facing endpoint returning at most 100 records, and it was hydrating the entire jobs directory on every request — #415's own pathology, in our own dashboard. `total` now comes from `ListJobsResult.total` (the pre-pagination match count #415 added for exactly this) rather than the length of a list that is no longer the full result.

## Measured

Against an isolated production-scale copy (2,016 records, 47.4 MB of `.yaml`), warm, two reps:

| call | time |
|---|---|
| `listJobs(dir)` + JS filter + `.slice(200)` | **1,935–2,095 ms** |
| `listJobs(dir, { agents, limit: 200 })` | **125–146 ms** |

**~15×.**

> The absolute numbers run about 2× higher than the 1,106 ms recorded in #418 — that measurement was taken on a quieter box. The ratio is what reproduces.

## Correctness

The downstream caller cares about **order**, not just membership, so equivalence is asserted rather than assumed:

- On the **real corpus**: returned ids and their order diffed against the JS-filtering shape — identical for n=200.
- In a **unit test**: four agents interleaved across 24 timestamps so that any per-agent grouping would reorder the result, then `{ agents, limit }` diffed against `listJobs(dir)` + `.filter()` + `.slice()`.

Also covered: empty array matches nothing, unknown agents ignored, AND-composition with `agent` and with `status`, and `total` as the pre-pagination count.

## Note for callers

`agents` **alone is not enough**. With `limit` still undefined, `retain = matches` and every match is hydrated regardless. Callers must pass **both**. The `listJobs` doc comment now says so with the numbers, since the failure mode is silent — you get correct results at full cost.

## Deliberately out of scope

Both named in #418:

- **`applyRetention`** genuinely wants every record — correct as-is.
- **Deleting `buildAttributionIndex`** is a breaking change to a public export with ~25 tests and a mock anchor. It deserves its own decision and probably a deprecation cycle, rather than riding along with a perf fix.

## Tests

- `packages/core` `job-metadata.test.ts`: **47 passed** (7 new)
- `packages/web` `routes.test.ts`: **42 passed** (2 new; 3 updated — they asserted the old slice-after-the-fact shape, so the mock now behaves like the real `listJobs` and applies `offset`/`limit` itself)
- `packages/core` builds clean.

## Summary by CodeRabbit

* **New Features**
  * Added filtering for jobs associated with multiple agents.
  * Combined agent filters are supported, including empty-list behavior.

* **Performance**
  * Improved job listing pagination by applying limits and offsets earlier.
  * Job totals now reflect all matching results, not just the current page.

* **Bug Fixes**
  * Improved pagination behavior alongside job filters.
  * Added safeguards for maximum page sizes and default pagination.

### DIFF
--- a/.changeset/feat-418-listjobs-agents-filter.md
+++ b/.changeset/feat-418-listjobs-agents-filter.md
@@ -4,8 +4,8 @@
 
 feat(state): filter `listJobs` by a SET of agents, and page the dashboard's `/api/jobs` inside it
 
-#415 made `listJobs` index-backed, but only for callers that can express what
-they want as a filter. A caller that needs several agents' jobs had no way to
+Issue `#415` made `listJobs` index-backed, but only for callers that can express
+what they want as a filter. A caller that needs several agents' jobs had no way to
 say so — `ListJobsFilter` had `agent` (exactly one) and nothing else — so it had
 to list the directory unfiltered and filter in JS, which reads, parses and
 Zod-validates every record in the directory to return a handful.
--- a/.changeset/feat-418-listjobs-agents-filter.md
+++ b/.changeset/feat-418-listjobs-agents-filter.md
@@ -1,0 +1,46 @@
+---
+"@herdctl/core": minor
+---
+
+feat(state): filter `listJobs` by a SET of agents, and page the dashboard's `/api/jobs` inside it
+
+#415 made `listJobs` index-backed, but only for callers that can express what
+they want as a filter. A caller that needs several agents' jobs had no way to
+say so — `ListJobsFilter` had `agent` (exactly one) and nothing else — so it had
+to list the directory unfiltered and filter in JS, which reads, parses and
+Zod-validates every record in the directory to return a handful.
+
+`ListJobsFilter` gains **`agents?: string[]`**. It combines with `agent` as an
+AND, and an empty array matches nothing (rather than being treated as "no
+filter", which would silently turn "I have no agents to look up" back into a
+full directory scan). Measured on a 2,016-record, 47.4 MB jobs directory, warm:
+
+| call | time |
+| --- | --- |
+| `listJobs(dir)` + JS filter + `.slice(200)` | 1,935–2,095 ms |
+| `listJobs(dir, { agents, limit: 200 })` | 125–146 ms |
+
+**~15×**, and the results are identical — the returned ids and their order were
+diffed against the JS-filtering shape on that real corpus, and in a unit test
+that interleaves four agents across time so any per-agent grouping would
+reorder.
+
+Note that `agents` alone is not enough to get this win. With `limit` undefined
+the index retains every match, so callers must pass **both** `agents` and
+`limit`; the `listJobs` docs now say so, with the numbers.
+
+**`packages/web`'s `/api/jobs` now pages inside `listJobs`** instead of asking
+for everything and slicing. It is a paginated, user-facing endpoint that
+returns at most 100 records, and it was hydrating the entire jobs directory on
+every request — the exact pathology #415 set out to remove, in our own
+dashboard. `total` now comes from `ListJobsResult.total` (the pre-pagination
+match count, added by #415 for precisely this) rather than from the length of a
+list that is no longer the full result.
+
+Unblocks [paddock#535](https://github.com/edspencer/paddock/issues/535), whose
+`listRunsForAgents` is the same bug downstream.
+
+The remaining unfiltered callers named in #418 are deliberately left alone here:
+`applyRetention` genuinely wants every record, and deleting the caller-less
+public `buildAttributionIndex` is a breaking change that deserves its own
+decision rather than riding along with a perf fix.
--- a/packages/core/src/state/__tests__/job-metadata.test.ts
+++ b/packages/core/src/state/__tests__/job-metadata.test.ts
@@ -417,6 +417,125 @@ describe("listJobs", () => {
     });
   });
 
+  describe("filtering by agents (set)", () => {
+    it("returns jobs for every listed agent and nothing else", async () => {
+      await createJob(tempDir, { agent: "agent-a", trigger_type: "manual" });
+      await createJob(tempDir, { agent: "agent-b", trigger_type: "manual" });
+      await createJob(tempDir, { agent: "agent-c", trigger_type: "manual" });
+
+      const result = await listJobs(tempDir, { agents: ["agent-a", "agent-c"] });
+
+      expect(result.jobs).toHaveLength(2);
+      expect(result.jobs.map((j) => j.agent).sort()).toEqual(["agent-a", "agent-c"]);
+    });
+
+    it("treats an EMPTY array as matching nothing, not as no filter", async () => {
+      // The alternative reading — empty means unfiltered — would silently turn a
+      // caller's "I have no agents to look up" into "hydrate the whole directory",
+      // which is the exact pathology this filter exists to remove.
+      await createJob(tempDir, { agent: "agent-a", trigger_type: "manual" });
+
+      const result = await listJobs(tempDir, { agents: [] });
+
+      expect(result.jobs).toEqual([]);
+      expect(result.total).toBe(0);
+    });
+
+    it("ignores listed agents that have no jobs", async () => {
+      await createJob(tempDir, { agent: "agent-a", trigger_type: "manual" });
+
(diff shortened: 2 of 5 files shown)


edspencer (commented):
Adds `agents?: string[]` to `ListJobsFilter` (ANDs with `agent`, empty array matches nothing) and switches the web `/api/jobs` route to page inside `listJobs` instead of hydrating the whole directory and slicing. The change is correct and thoroughly tested — no blocking issues found.

edspencer:
I verified `ListJobsResult.errors` is typed `number` in `packages/core/src/state/job-metadata.ts`, so the prior `errors: []` in the route tests was an incorrect shape and `errors: 0` is the accurate fix — the route forwards `result.errors` verbatim, so mocks should model a count. The `ListJobsFilter` import/annotation on the pagination mock is a strictness improvement over `any` and reads `filter?.offset`/`filter?.limit` consistently with the real implementation. The changeset change only reflows the `#415` reference into prose. This is a test-only + docs change with no runtime surface; nothing to flag.

Reviewed 2 changed files (8 hunks); 0 findings.

edspencer:
Thanks — addressed three of the four in 8da4943:

- **`errors` as an array in the `/api/jobs` mocks** — good catch, and it was pre-existing throughout the block rather than just my new cases. `listJobs` returns `errors` as a *number*, so those tests were asserting against a response shape production never emits. Fixed all six.
- **`filter: any` in the paging mock** — fair; a test whose job is to notice filter-contract changes shouldn't opt out of the contract. Now `ListJobsFilter`.
- **MD018 in the changeset** — fixed.

**Declining the "mock the filesystem boundary" suggestion on `job-metadata.test.ts`**, with reasoning rather than just a no:

Every test in that file is filesystem-backed by design — it's built on `createTempDir` + `writeJobFile`, and `listJobs`'s whole contract is *"filter and page against an mtime-keyed index of files on disk"*. The index, the mtime/size cache key, and the `retain` rule being exercised **are** the filesystem boundary; there is no meaningful unit of this function above it.

More specifically, the load-bearing new test asserts that filtering through the real index returns **the same records in the same order** as the `listJobs(dir)` + `.filter()` + `.slice()` shape it replaces — that equivalence is the entire reason the downstream change (edspencer/paddock#535) is safe. Substituting constructed index records for real files would mean the test could no longer fail if the real index diverged, which is the only failure it exists to catch.

Happy to revisit if there's an existing in-memory storage fake in core I've missed.

edspencer (commented):
Follow-up commit that corrects test-mock fidelity for the `listJobs`/`/api/jobs` change: the `/api/jobs` route mocks now return `errors: 0` (a `number`) instead of `errors: []`, matching the real `ListJobsResult.errors: number` type, and the pagination mock is typed with `ListJobsFilter` rather than `any`. The changeset edit is cosmetic prose. No issues found.

Issue #263, opened 2026-06-21
Runtime-added agent can't resume a session created by another agent in the same process (until restart)


edspencer:
## Summary

When an agent (B) is asked to resume a Claude Code session that was **created by a different agent (A) in the same long-running FleetManager process**, the CLI runtime forks a fresh session instead of continuing — **even though the transcript has been relocated into B's working directory, the `cwd` rewritten, and a job record re-attributes the session to B.** A process restart fixes it; nothing at runtime does.

Found while building paddock's "promote a one-off chat into a project" feature (move a scratch-agent chat to a project keeper).

## Repro (paddock context, but the mechanism is core)

1. Agent `scratch` (cwd `/data/scratch`) runs a turn → creates session `X` (transcript at `~/.claude/projects/<enc(/data/scratch)>/X.jsonl`).
2. Move `X.jsonl` → `~/.claude/projects/<enc(/data/projects/foo)>/X.jsonl` (keeper `keeper-foo`'s cwd), rewrite the embedded `"cwd"` to `/data/projects/foo`.
3. Write/repoint job record(s) so the attribution index maps `X → keeper-foo`.
4. `fleet.invalidateSessions("keeper-foo")` + `invalidateSessions("scratch")`.
5. `getAgentSessions("keeper-foo")` now lists `X` ✅ and `getAgentSessionMessages` hydrates it ✅.
6. `trigger("keeper-foo", { resume: "X", prompt })` → **forks a brand-new session; the model has no prior context.** ❌

## What's been ruled out

- **The transcript/cwd/symlink are correct.** Running `claude --resume X -p "…"` directly in `/data/projects/foo` resumes perfectly and recalls prior context. ✅
- **Not the attribution/discovery cache.** Listing + history work immediately after `invalidateSessions`.
- **Not timing.** A 6 s delay before resume still forks.
- **Not the agent being runtime-added per se.** A *native* runtime-added project whose keeper **created** its own session resumes fine at runtime. ✅ The break only happens when the session was created by a **different** agent in this process.
- **Re-registering** the keeper (`addAgent` replace) and **`deleteSession(scratch, X)`** (to evict the creator's tracking) do **not** fix it.
- **A full process restart fixes it** — after restart, `keeper-foo` resumes `X` with full continuity. ✅

## Diagnosis

[...]

PR #433, opened 2026-08-01
fix(core): resolve the Claude home consistently, and tell Claude Code about it (#423)

edspencer:
# Resolve the Claude home consistently, and tell Claude Code about it

Part 1 of 2 for #423. This is a **standalone bug fix**: it matters to anyone
running herdctl against a non-default Claude home, whether or not they care
about session adoption. The adoption feature itself is in the follow-up PR,
which stacks on this one.

## The problem: three names for one concept

The "Claude home" is the `.claude` directory that holds
`projects/<encoded-cwd>/<session-id>.jsonl` transcripts. Three separate layers
each have their own name for it, and nothing was making them agree:

| Name | Owned by | Where it appears |
|------|----------|------------------|
| `CLAUDE_HOME` (or similar) | the embedding app | e.g. paddock resolves one and launches Claude against it |
| `claudeHomePath` | herdctl | `SessionDiscoveryOptions` — and, before this PR, nowhere else |
| `CLAUDE_CONFIG_DIR` | Claude Code | the env var the Agent SDK and the `claude` binary read to find their own home |

Two distinct bugs fall out of that.

## Bug 1 — the listing path and the read path disagreed

`SessionDiscoveryService` already accepted an injectable `claudeHomePath` and
scanned `<claudeHome>/projects/` for transcripts. But `getCliSessionDir()` and
`getCliSessionFile()` hardcoded `path.join(os.homedir(), ".claude")`.

So under a non-default home, discovery **listed** sessions out of the configured
home and then **read** each one out of `~/.claude`, where nothing was. Sessions
listed but opened empty.

Both helpers now take an optional trailing `claudeHomePath`, falling back to a
new exported `defaultClaudeHome()` — a function rather than a module constant, so
`os.homedir()` is read at call time. `FleetManager` resolves the home once in its
constructor, exposes it via `getClaudeHomePath()`, and threads it into session
discovery, `RuntimeFactory`, `SDKRuntime`, `CLIRuntime`, `JobControl`,
`ScheduleExecutor`, `runSchedule()`, `deleteSession()` and
`cliSessionFileExists()`.

This bug is completely invisible whenever the configured home happens to equal
`~/.claude`, which is why it lurked.

## Bug 2 — Claude Code resolves its own home from `CLAUDE_CONFIG_DIR`, and nothing set it

**This is the part that will look like scope creep if you skim it, so please read
this section.** An environment variable appearing in a path-resolution fix is not
incidental — it is the half of the bug that threading alone cannot fix.

Threading `claudeHomePath` fixes herdctl's *own* path arithmetic. But the process
that actually **writes** transcripts is Claude Code — the Agent SDK for the `sdk`
runtime, the spawned `claude` binary for the `cli` runtime. Neither has a "Claude
home" option to pass. Both resolve their home from the `CLAUDE_CONFIG_DIR`
environment variable, and herdctl was never setting it.

Left unset, herdctl and Claude Code operate on **different trees**.

### Evidence (verified live, not reasoned about)

Both of these were reproduced on a running instance with a non-default home:

1. **New chats landed in the wrong tree.** Started a fresh chat through herdctl.
   Its transcript appeared under `~/.claude/projects/…`, while herdctl watched
   and listed `<claudeHomePath>/projects/…` and saw nothing appear. Silent — no
   error, the chat just never showed up where herdctl looks.

2. **Resuming failed outright.** Took a session whose transcript lived in the
   configured home (exactly the case adoption produces) and resumed it. The turn
   died with `error_during_execution`: Claude Code was handed a session id whose
   file it could not find, because it was looking under `~/.claude`.

3. **Setting `CLAUDE_CONFIG_DIR` fixed both, live.** New chats landed in the
   configured home; resume worked.

### How it's applied

New module `packages/core/src/runner/runtime/claude-config-dir.ts` exports
`CLAUDE_CONFIG_DIR_VAR`, `resolveClaudeConfigDir()` and `withClaudeConfigDir()`.

- **`SDKRuntime`** applies the variable to the per-query `sdkOptions.env`, as the
  last step of building its options. Scoped to the query rather than mutating
  `process.env`: a host runs many concurrent agents, and a global mutation would
  leak one agent's home into all of them. Note the SDK's `env` **replaces** the
  subprocess environment wholesale rather than merging, which is why
  `withClaudeConfigDir()` spreads the inherited environment itself.
- **`CLIRuntime`** adds it to its default `execa` spawn. `execa` merges `env` over
  the inherited environment (`extendEnv` defaults to true), so it's a per-spawn
  addition. A caller-supplied `processSpawner` owns its own env and is left alone.
- **`ContainerRunner` deliberately injects nothing.** The container has its own
  filesystem and its own fixed home: `HOME=/home/claude`, with
  `/home/claude/.claude/projects/-workspace` bind-mounted back to the host
  `<stateDir>/docker-sessions` — which is how herdctl reads those transcripts at
  all. A host path is meaningless in there and would move the agent's transcripts
  off the mount and out of herdctl's view: the same split-brain, inverted. There's
  a comment saying so at the injection site.

Two escape hatches, both no-ops for existing users:

- **The default home injects nothing.** `resolveClaudeConfigDir()` returns
  `undefined` for `~/.claude`, so anyone not using a custom home sees byte-identical
  behaviour.
- **An operator who already set `CLAUDE_CONFIG_DIR` wins.** herdctl never
  overwrites it.

## API surface (all additive, hence `minor`)

- `FleetManagerOptions.claudeHomePath`, `FleetManager.getClaudeHomePath()`
- `SessionDiscoveryService.getClaudeHomePath()`
- `SDKRuntimeOptions` (with `claudeHomePath`), `SDKRuntime.getClaudeHomePath()`,
  `CLIRuntime.getClaudeHomePath()`
- `RuntimeFactory.create()` accepts `claudeHomePath`
- `RunScheduleOptions.claudeHomePath`, `SessionFileCheckOptions.claudeHomePath`,
  third `claudeHomePath` parameter on `cliSessionFileExists()`
- `defaultClaudeHome()`, `CLAUDE_CONFIG_DIR_VAR`, `resolveClaudeConfigDir()`,
  `withClaudeConfigDir()`
- `SDKQueryOptions.env`, mirroring the SDK's own `Options["env"]`
- `FleetManagerContext.getClaudeHomePath?()` — **optional** on purpose, so the
  lightweight mock contexts the module unit tests build don't all need updating;
  callers fall back to the path helpers' own default, which is exactly the
  pre-#423 behaviour

Every new parameter is optional and defaults to the previous behaviour.

## Tests

Four new suites, all exercising the real threading rather than mocks of it:

- `state/__tests__/claude-home-threading.test.ts` — discovery resolves reads
  against the configured home
- `runner/runtime/__tests__/claude-home-threading-cli-runtime.test.ts` —
  `CLIRuntime` path math and home resolution
- `runner/runtime/__tests__/claude-config-dir-threading.test.ts` — the
  `CLAUDE_CONFIG_DIR` injection: default home injects nothing, an operator's own
  value wins, `process.env` is never mutated
- `fleet-manager/__tests__/claude-home-threading-job-control.test.ts` — the
  `JobControl` context wiring, including the `SDKRuntime` resume path

`cli-runtime.test.ts` also gains a `defaultClaudeHome` stub, because it mocks the
path-helpers module wholesale.

## Review notes

- `SessionDiscoveryService.getClaudeHomePath()` carries a doc comment mentioning
  session adoption as a motivating caller. That's the follow-up PR; the accessor
  stands on its own here (it's what lets any caller avoid re-deriving
  `os.homedir()/.claude`) and is covered by `claude-home-threading.test.ts`.
- The `FleetManagerContext.getClaudeHomePath?()` optionality is a deliberate
  compatibility choice, not an oversight — see above.

## Verification

- `pnpm --filter @herdctl/core build` — clean, module loads
- Full core sweep: **3546 passed, 1 skipped, 1 failed** — the single failure is
  the pre-existing `state/__tests__/directory.test.ts > "throws
  StateDirectoryCreateError when parent directory is not writable"`, which fails
  on any box running tests as root (a chmod-restricted dir is still writable) and
  is unrelated to this change.

Closes part 1 of #423.

## Summary by CodeRabbit

* **New Features**
  * Added configurable Claude home directory support across session discovery, chat, scheduled jobs, and CLI/SDK runtimes.
  * Added consistent transcript and session resolution for custom Claude home locations.
  * Added environment configuration handling that preserves explicit settings and avoids host environment mutation.
  * Added public accessors and configuration options for resolving the active Claude home.

* **Bug Fixes**
  * Improved consistency when resuming sessions and locating transcripts across runtime types.
  * Preserved container-specific Claude configuration behavior.

* **Documentation**
  * Documented Claude home configuration, transcript resolution, and runtime behavior.


edspencer (commented):
This PR resolves the Claude home (`.claude` transcript directory) consistently across herdctl's own path arithmetic **and** tells the Claude Code process about it via `CLAUDE_CONFIG_DIR`. It threads an optional `claudeHomePath` from `FleetManager` through session discovery, `RuntimeFactory`, `SDKRuntime`, `CLIRuntime`, `JobControl`, `ScheduleExecutor`, `runSchedule()`, `deleteSession()` and `cliSessionFileExists()`, adds a new `claude-config-dir` module (`resolveClaudeConfigDir`/`withClaudeConfigDir`), and injects the env var per-query (SDK) / per-spawn (CLI) without ever mutating `process.env`.

I read the new `claude-config-dir.ts` module in full, `sdk-runtime.ts` (`buildSdkOptions`/`execute`/`openSession`), `cli-runtime.ts`, `factory.ts`, `container-runner.ts`, and the session-discovery threading, and cross-checked the four new test suites. I verified: (1) the SDK and CLI injection paths both go through `buildSdkOptions`/the default spawner and correctly no-op for the default home or an operator-set value; (2) the SDK `env` copy spreads the inherited environment (since the SDK replaces rather than merges); (3) `ContainerRunner` genuinely bypasses the wrapped runtime's spawner (its own `executeCLIRuntime`/`executeSDKRuntime` build docker-exec env directly, so no host path leaks in); and (4) `getCliSessionDir`/`getCliSessionFile` fall back to `defaultClaudeHome()` at call time. No correctness, security, or async issues found — the change is additive, backward-compatible, and the escape hatches (default home / operator-set var) are covered by tests.

edspencer:
Reviewed 9 changed files (14 hunks); 0 findings.

edspencer (commented):
This incremental change extends the #423 Claude-home threading to the **resume-fallback / session-validation** path, which was the remaining gap: the CLI-runtime transcript existence check that `getSessionInfo` runs (via `validateSessionWithFileCheck`) is destructive — it CLEARS a session it judges stale — so a non-default home previously destroyed valid session pointers. It adds `SessionOptions.claudeHomePath` (consulted only for `runtime: "cli"`), an optional `getClaudeHomePath?()` on `RuntimeInterface`, and threads the home in at the three validation call sites (`job-executor.ts`, `job-control.ts`, `schedule-runner.ts`) plus the `cliSessionFileExists` adopt check.

I read `job-executor.ts` (both the timeout-aware `getSessionInfo` read and the adopt branch), `session.ts` `getSessionInfo`, and `session-validation.ts` `validateSessionWithFileCheck`. I verified: (1) the Docker path is unaffected — `validateSessionWithFileCheck` routes `docker_enabled` sessions to `dockerSessionFileExists` and never consults `claudeHomePath`, and the adopt branch is guarded by `!dockerEnabled`; (2) `this.runtime.getClaudeHomePath?.()` returns the same home `RuntimeFactory` gave the runtime, so the existence check and the runtime's actual reads/writes agree by construction (`ContainerRunner` correctly returns `undefined` → default, but its sessions never reach the CLI file check); (3) SDK runtime skips the file check entirely, so threading is inert there; and (4) `schedule-runner` already had `claudeHomePath` in scope from part 1. The new tests exercise the real (unmocked) validation and assert both the fix and the destructive control case. No correctness, security, or async issues found — additive and backward-compatible.

edspencer:
Addressed the three review findings in `3e8abf5` — but finding 1 turned out to be wider than reported, so recording that here.

**The root cause was upstream of all three call sites.** `SessionOptions` had no field to carry a Claude home, so `getSessionInfo` always called `validateSessionWithFileCheck` without one. Patching only the three named sites would have compiled and fixed nothing.

**Reachability of the sites differs from the review.** Checking which actually run the file check:

| Site | Live today |
|---|---|
| `schedule-runner.ts:322` (reported) | yes |
| `job-control.ts:167` (reported) | no — passes no `runtime`, so the check never runs. Threaded anyway, correct-but-inert |
| `job-control.ts:387` (reported) | no — streaming sessions are always SDK runtime |
| **`job-executor.ts:233`** (not reported) | **yes** — passes `runtime`, so it fires for `cli` agents |
| **`job-executor.ts:357`** (not reported) | **yes** — the #263 CLI adoption probe, unconditional, no home argument at all |

So one of the three reported sites was reachable, and the two most live ones were not in the review.

`job-executor` reaches the home via a new optional `getClaudeHomePath?()` on `RuntimeInterface`; both first-party runtimes implement it, and `ContainerRunner` correctly leaves it undefined since its home lives inside the container.

Ruled out as genuinely inert rather than threaded defensively: `job-executor.ts:231`/`:791`, `cleanupExpiredSessions`, and the two `packages/cli` callers all pass no `timeout`, so the validation block is skipped. A repo-wide sweep confirms every `getCliSessionFile` / `getCliSessionDir` / `cliSessionFileExists` call now threads a home.

**The consequence is worse than a failed resume.** `getSessionInfo` *deletes* a session it judges stale — so under a non-default home a valid session was destroyed and the run silently restarted from scratch, rather than merely being skipped.

New coverage: a `getSessionInfo` block in `claude-home-threading.test.ts`, and a new `claude-home-threading-schedule-runner.test.ts` driving `runSchedule` end-to-end with a `cli` agent and a real transcript in an alternate home (exercising the schedule-runner and job-executor sites in one chain). Both verified red-before with their controls passing.

PR #434, opened 2026-08-01
feat(core): adopt pre-existing Claude Code CLI sessions into an agent (#423)

### LINKED ISSUE #423: Adopt pre-existing Claude Code CLI sessions into an agent

edspencer:
# Adopt pre-existing Claude Code sessions into an agent

Part 2 of 2 for #423.

> **Depends on #433 (`fix/423-claude-home-resolution`).** This branch is stacked
> on it — please review and merge that one first. Adoption places transcripts
> into, and resumes them out of, the fleet's configured Claude home, so it is
> only correct once the home resolves consistently and Claude Code is told about
> it via `CLAUDE_CONFIG_DIR`. Reviewing this PR against `main` will show A's diff
> as well; diff against the base branch instead.

## The problem

A Claude Code session a user ran themselves in a terminal is visible in
all-sessions views but invisible under any agent, because nothing attributes it
to one. There was no way to say "this pile of chats I already have belongs to
this agent".

### Adoption is a dedicated store, not a forged job record

Records live at `<stateDir>/adopted-sessions/<session-id>.yaml`:

```yaml
version: 1
sessionId: 0f3c…
agentName: my-fleet/my-agent
adoptedAt: 2026-08-01T12:00:00.000Z
sourceCwd: /Users/ed/Code/myproject
```

The tempting shortcut is to write a synthetic job record, since the attribution
index already reads those. It's the wrong shape: **a job record asserts that a
run happened.** Forging one to buy attribution would make job listings, history
and any metric derived from them lie about work that never executed, and it
would be indistinguishable from a real run forever after. A dedicated store says
what it means.

The directory is created lazily on first write, like the sparse
`<platform>-sessions` stores; a missing directory just means "no adoptions".
Session IDs arrive from user input (CLI args, HTTP bodies), so file paths go
through `buildSafeFilePath()` — a hostile id like `../../etc/passwd` throws
rather than escaping the store.

### Precedence: job → platform → adopted → native

`buildAttributionIndex()` (and the incremental `AttributionIndexBuilder`) now
read a third source, and `SessionOrigin` gains `"adopted"`.

Adoption is checked **last**, immediately before the native fallback. A real run
record or a live platform binding is stronger evidence of where a session came
from than an after-the-fact claim, so adoption never overrides them — it only
rescues sessions that would otherwise be unattributed. `adopted` also has no
trigger type; it doesn't come from a job.

Adoption records are few and mutable (a session can be adopted, re-adopted by
another agent, or released), so unlike job records they are re-read in full on
each build rather than mtime-cached. Same treatment platform session files
already get.

### Copy by default, with mtime preservation

Discovery looks for an agent's transcripts under
`<claudeHome>/projects/<encoded agent cwd>/`, so a transcript recorded under a
*different* directory isn't where discovery will look. `adoptSessionsFrom()`
therefore places it into the agent's own folder, then records attribution with
the originating directory as `sourceCwd`. When the source already resolves to the
agent's own folder, nothing is moved — attribution only.

| Mode | Effect |
|------|--------|
| `copy` (**default**) | Duplicate, preserving the source mtime. The user's original `~/.claude` transcript is untouched |
| `move` | Relocate. It disappears from the user's terminal history |
| `link` | Hard-link (symlink across devices). One inode, so a resume appends to the user's original too |

`copy` is the default because the alternative is mutating files the user did not
ask us to touch. **Copies preserve the source mtime deliberately**: mtime drives
both list ordering and every metadata cache key, so stamping a fresh one would
reorder the user's whole history and cold-start their caches. Existing
destination files are never overwritten.

Every non-adopted candidate lands in `skipped` with a reason a UI can show
verbatim — `sidechain`, `already-adopted`, `destination-exists`,
`attributed-to-run`, `unreadable`, `placement-failed`, `record-failed` — and one
bad transcript never aborts the batch. `dryRun: true` performs **no** writes at
all and reports what would have happened.

Placement targets the CLI transcript folder, so batch adoption is for non-Docker
agents. A Docker-wrapped agent reads its sessions from
`<stateDir>/docker-sessions/` (the container's `~/.claude` is ephemeral) and
should claim them with `adoptSession()`, which records attribution without moving
anything.

### Title precedence by entry type, not file position

Adopted terminal sessions rendered as their raw session id, because auto-naming
read only `type: "summary"` entries — and CLI transcripts essentially never emit
one (those come from herdctl-driven runs).

New `extractSessionTitle()` returns the highest-precedence title present:

1. `custom-title`, value in a **`customTitle`** field — the user set it, so it wins
2. `ai-title`, value in an **`aiTitle`** field — Claude Code's own generated title
3. `summary`, value in a `summary` field

Precedence is by **entry type, not by file position**: a later `ai-title` must
never clobber an earlier `custom-title`, which a naive last-wins scan would do.
Within a single type the last occurrence wins, because titles get rewritten as a
session evolves — hence streaming to EOF with no early exit, matching
`extractLastSummary()`. An entry carrying a plain `title` field is not a title we
understand and is ignored (the field names really are `customTitle`/`aiTitle`).

`resolveAutoName` falls back to the first user message when a transcript has no
title at all. `extractLastSummary()` is left untouched so the contract of
`extractSessionMetadata()`, which also reads summaries, is unchanged.

### `autoNameVersion` is an optional field, not a file-schema bump

This one is subtle and worth a close look.

The auto-name cache is authoritative on the presence of `autoNameMtime`, **not**
of `autoName` — a transcript that yields no name is negative-cached so it's never
re-streamed. That means changing the extractor alone would have done *nothing* to
existing data: every already-listed session holds an entry whose mtime is still
current, produced by the old extractor, so the old (usually empty) result would
keep winning forever. The feature would silently not ship.

So `SessionMetadataEntrySchema` gains an optional `autoNameVersion`, stamped by
`setAutoName`/`batchSetAutoNames` with `AUTO_NAME_EXTRACTOR_VERSION`. An entry
whose version doesn't match is treated as a miss, re-extracted with current logic
and rewritten stamped — once per session, not once per listing — while
`customName`, `preview`, `usage` and `isSidechain` in that entry survive
untouched.

**Why not bump `SessionMetadataFileSchema`'s `version`?** Because
`loadMetadata()` discards the **entire file** when it fails to parse. A file
version bump would make every existing metadata file fail validation and be
thrown away, silently destroying every user-set `customName` — a data-loss event
dressed up as a cache invalidation. The field is optional for the same reason: a
file written by an older release must still validate.

## New API

| Method | Purpose |
|--------|---------|
| `listAdoptableSessions(name, fromWorkingDir?)` | Candidates, newest first — backs an "import my existing chats" picker |
| `adoptSession(name, sessionId, opts?)` | Claim one session by id, moving nothing. Idempotent |
| `adoptSessionsFrom(name, opts?)` | Place and claim every adoptable session in a directory |
| `unadoptSession(name, sessionId)` | Release this agent's claim; the transcript stays on disk |

`AdoptableSession` is deliberately **not** a `DiscoveredSession`: half that shape
is meaningless before adoption. `origin` is by definition `native`, `agentName`
by definition undefined (that's *why* it's invisible), `resumable` is a property
of the adopting agent rather than the candidate, and `customName` is keyed per
agent. So it's `sessionId`, `sourceCwd`, `mtime`, `autoName`, `preview`,
`sizeBytes`. No message count: obtaining one means streaming the whole transcript,
exactly the per-session cost the listing caches exist to avoid. `sizeBytes` comes
free from the `stat()` the directory listing already does.

`unadoptSession()` returns `false` — rather than removing anything — when the
session isn't adopted, or when it's adopted by a *different* agent. One agent must
not be able to drop another's claim.

Also exported: `ADOPTED_SESSIONS_DIR_NAME`, `getAdoptedSessionsDir`,
`getAdoption`, `listAdoptions`, `recordAdoption`, `removeAdoption`,
`RecordAdoptionOptions`, `ADOPTED_SESSION_VERSION`, `AdoptedSession`,
`AdoptedSessionSchema`, `AdoptedSessionVersionSchema`, `extractSessionTitle`,
`AUTO_NAME_EXTRACTOR_VERSION`, and the `AdoptionPlacementMode`,
`AdoptSessionsFromOptions`, `AdoptSessionsResult`, `AdoptSkippedSession`,
`AdoptSkipReason` types.

## Tests

- `state/__tests__/adopted-sessions.test.ts` — the store: CRUD, path safety,
  lazy directory creation, schema validation
- `state/__tests__/session-adoption.test.ts` — the full adoption surface:
  placement modes, mtime preservation, every skip reason, dry run, cross-agent
  unadopt refusal
- `state/__tests__/session-adoption-metadata-refusal.test.ts` — adoption survives
  an unreadable metadata file (the #419 refusal path)
- `state/__tests__/auto-name-extraction.test.ts` + four `title-*.jsonl` fixtures —
  title precedence, including `title` vs `customTitle` field naming and
  last-within-type
- Additions to `session-attribution.test.ts` (third source, precedence),
  `jsonl-parser.test.ts` (`extractSessionTitle`), `session-metadata.test.ts`
  (version stamping, legacy entries still parse), and `session-discovery.test.ts`
  (legacy entry re-extracted exactly once; first-message fallback)

## Verification

- `pnpm --filter @herdctl/core build` — clean, module loads
- Full core sweep: **3622 passed, 1 skipped, 1 failed** — the single failure is
  the pre-existing `state/__tests__/directory.test.ts > "throws
  StateDirectoryCreateError when parent directory is not writable"`, which fails
  on any box running tests as root and is unrelated.

Closes #423.

## Summary by CodeRabbit

* **New Features**
  * Added support for adopting existing Claude Code sessions.
  * Sessions can be previewed, adopted individually or in batches, and placed by copying, moving, or linking.
  * Added dry-run support, per-session skip reasons, collision handling, and the ability to release adopted sessions.
  * Improved automatic session naming using custom titles, AI-generated titles, summaries, and user-message fallbacks.

* **Bug Fixes**
  * Improved resilience when session metadata is missing or unreadable.
  * Added cache invalidation to ensure updated session names are discovered correctly.

* **Documentation**
  * Documented session adoption workflows, attribution, placement modes, and public APIs.


packages/core/src/state/session-discovery.ts
```diff
@@ -1260,6 +1408,457 @@ export class SessionDiscoveryService {
...
+    //
+    // Wrapped in persistCacheUpdates so a metadata file that the store refuses
+    // to overwrite (issue #419) degrades to a cold cache rather than aborting
+    // the scan — and with it both `listAdoptableSessions` and
+    // `adoptSessionsFrom`, which would otherwise fail to adopt anything because
+    // one unrelated cache file is corrupt.
+    if (sidechainUpdates.length > 0) {
+      await this.persistCacheUpdates(() =>
+        this.sessionMetadataStore.batchSetSidechains(agentName, sidechainUpdates),
+      );
```
edspencer: **LOW · correctness** — dryRun adoption still writes the sidechain metadata cache, contradicting the documented "no writes at all"

`scanAdoptionCandidates()` unconditionally persists freshly-computed sidechain flags via `batchSetSidechains`, and it is called by `adoptSessionsFrom()` even when `dryRun: true`. So a dry run of adoption can create or modify `session-metadata/<agent>.json` on disk whenever any candidate needs sidechain classification (e.g. the `dryRun` test's `sess-side` triggers exactly this path).

This conflicts with the contract asserted repeatedly in the docs/changeset and the method's own JSDoc: "With `dryRun: true` nothing at all is written" / "Perform no writes at all". The `session-adoption.test.ts` dryRun case only asserts that no transcript was placed and no adoption record was written — it does not check the metadata dir — so the deviation is untested.

The inline comment acknowledges this is intentional and harmless (it's a pure cache that memoizes a fact already true on disk), which is a fair position. But a caller relying on `dryRun` for a genuinely side-effect-free preview (e.g. read-only inspection) would be surprised to see a metadata file appear. Either gate the sidechain persist on `!dryRun` (threading a flag into `scanAdoptionCandidates`), or soften the "no writes at all" wording to "no transcript placement and no adoption record" so the documented contract matches the code.

edspencer (commented):
Solid, well-tested PR. It threads a configurable Claude home end-to-end (discovery, path helpers, both runtimes, and `CLAUDE_CONFIG_DIR` for the underlying Claude Code process), adds a dedicated adoption store as a third attribution source, and version-stamps the auto-name cache so a changed extractor actually invalidates existing entries.

What I checked: read the misrendered-as-binary `claude-config-dir.ts` in full (it's fine — real content, correct `resolveClaudeConfigDir`/`withClaudeConfigDir` logic, no `process.env` mutation); traced the adoption path in `session-discovery.ts` (`scanAdoptionCandidates`, `adoptSessionsFrom`, `placeTranscript`, cache invalidation) and confirmed the copy/move/link modes, never-clobber via `COPYFILE_EXCL`, EXDEV/EPERM fallbacks, and per-session skip resilience; verified the attribution precedence (job → platform → adopted → native) and that `invalidateWorkingDirectory`/`invalidateAttributionCache` both drop the attribution index so adopted sessions surface immediately; confirmed `listAdoptions`' throw-on-non-ENOENT matches the existing `buildPlatformIndex` behavior (not a regression); checked the `autoNameVersion` optional-field approach and negative-cache re-extraction logic. Only one minor contract discrepancy found (below).

edspencer:
Reviewed 46 changed files (137 hunks); ran the verify pass; 1 finding.

edspencer:
Addressed the review in `c7df3cf`. Two of the four were valid; recording the reasoning since one correction cuts both ways and one finding is declined.

**1. `dryRun` wrote the sidechain metadata cache — valid, fixed.** `scanAdoptionCandidates()` persisted freshly-computed sidechain flags unconditionally, and `adoptSessionsFrom()` called it regardless of `dryRun`. Proven rather than reasoned — asserting the whole state dir is empty after a dry run gave `expected [ 'session-metadata' ] to deeply equal []`. Gated the write rather than softening the docs: the "nothing at all is written" promise appears in four places and is exactly what a caller previews against a pristine or read-only state dir; all it buys is one re-read of each candidate's first line on the next real call. `listAdoptableSessions` still warms the cache.

**2. `move` could clobber — valid, but the framing is off in BOTH directions.**

*Narrower than implied:* the caller already pre-checks (`stat(destFile)` → `destination-exists`) for every mode, so the plain "destination already holds this transcript" case was safe. The test suggested in review passes **before** the fix; I added it as a caller-path guard but it is not the red one.

*Wider than reported, and this is the real bug:* the pre-check is check-then-act, so it misses anything a destination `stat()` fails on — deterministically, no race required. **A dangling symlink reads as ENOENT and is waved straight through.** Reachable from our own code: `mode: "link"` falls back to `symlink()` on EXDEV/EPERM, so a cross-device placement leaves exactly that behind once the user removes the original. `copy` survived by accident (`COPYFILE_EXCL` fails EEXIST even on a dangling symlink); **`move` consumed the user's only copy and reported success.** Red-before: `expected [ 'sess-ghost' ] to deeply equal []`.

Fixed as a class, not an instance: every mode now creates the destination exclusively, so never-clobber is enforced by the syscall and the pre-check is demoted to an optimisation that yields a tidier skip reason. `move` no longer uses `rename` at all — it hard-links then unlinks the source (same net effect, one inode, mtime carried for free), with EXDEV/EPERM falling back to copy-then-unlink. Swept for the same pattern elsewhere: the only other `rename` in core is `renameWithRetry` in `state/utils/atomic.ts`, where replacing the destination is the entire point.

**3. Ownership guard on `unadoptSession` — declined, the guard is real.** `fleet-manager.ts:1068` reads the record via `getAdoption` and returns `false` unless `existing.agentName === agent.qualifiedName`; the test at `session-adoption.test.ts:597` adopts as another agent, asserts `false`, and asserts the record survives still owned by that agent. Delete the guard and both assertions fail. The suggested scenario — two agents clearing the same `sessionId` concurrently both passing the read-time check — cannot occur: there is one record per session with one owner, so at most one agent passes. The genuine (much narrower) race is unadopt-racing-a-re-adopt-by-another-agent, and moving the check inside `removeAdoption` would not close it — that is the same check-then-act one stack frame lower, and POSIX has no "unlink if contents match". Adopting it would imply an atomicity we would not have.

**4. Empty-result case — documented, with a semantics question left open.** Confirmed: `adoptSessionsFrom` returns `{adopted: [], skipped: []}` when the agent has no `working_directory`, before `opts` is consulted. Now documented in the JSDoc and both doc pages, with the reasoning (placement needs the agent's own transcript folder as destination, and `getAgentSessions` is scoped to that same directory).

But the asymmetry is arguably on the other side: given `getAgentSessions` also returns `[]` for a directory-less agent, `adoptSessionsFrom`'s refusal is the consistent behaviour — it is `listAdoptableSessions`' fallback to `fromWorkingDir` that is the outlier, since it will populate a picker for an agent that then adopts nothing and returns an empty `skipped` too. Left as an API-semantics call rather than changed in a review pass; no caller is affected today.

Nitpicks: applied the lstat/nlink assertion (`stat` follows symlinks, so a symlink placement would have reported the source inode and passed), PathTraversalError coverage at the public API, the orphan-cleanup on a failed source unlink, and the `scanBaseDirectory` rename. Skipped the `z.iso.datetime()` migration (repo convention is `z.string().datetime()`; belongs in a repo-wide Zod-4 pass) and the single-read enrichment restructure (real regression risk on a negative-cached cold path).

Sweep: 3632 passing, one known pre-existing root-chmod failure. Biome identical to baseline.

PR #438, opened 2026-08-02
fix(core): adopt only unattributed sessions, never a sibling agent's (#437)

### LINKED ISSUE #437: Adoption offers sessions owned by another agent, then silently loses them
## Summary

`listAdoptableSessions` / `adoptSessionsFrom` offer sessions that are already
attributed to a **different agent**. Adopting them writes an adoption record,
reports success, and the session then **never appears** — because attribution
resolves job records ahead of adoptions, so the session still belongs to the
other agent. The adoption marker is burned, so the session is excluded from every
future attempt. Silent, unrecoverable, no error surface.

Introduced by #423 (adoption), shipped in `@herdctl/core@5.29.0`.

## Root cause — `origin: "native"` does not mean "unattributed"

`scanAdoptionCandidates` gates on **origin**:

```js
const attribution = attributionIndex.getAttribute(sessionId);
if (attribution.origin !== "native") { skipped.push({ reason: "attributed-to-run" }); continue; }
```

`getAgentSessions` gates on **agentName**:

```js
if (attribution.agentName !== agentName) return { session: null, ... };
```

Those two disagree, because `triggerTypeToOrigin` folds several trigger types
into `"native"`:

```js
// manual, webhook, chat, fork — all treated as native CLI usage
default: return "native";
```

So a session with a **real job record** whose `trigger_type` is `chat` (or
`manual`/`webhook`/`fork`) resolves to:

```
{ origin: "native", agentName: "sweeper-myproject", triggerType: "chat" }
```

`origin` is genuinely `"native"`, so the adoption filter lets it through. But
`agentName` is set to a *sibling* agent, so the listing gate hides it. The
adoption record is written and loses the precedence contest
(job → platform → adopted → native), so it can never win.

The filter is asking "is this native?" when it needs to ask "is this
**unattributed**?".

## Impact — measured on a real instance

QA on ~2441 real transcripts across 7 workspaces, where a keeper and its sweeper
share one transcript directory:

| Workspace | reported adopted | actually visible | silently lost |
|---|---|---|---|
| A | 1382 | 26 | **1356** |
| B | 96 | 2 | **94** |
| C | 13 | 0 | **13** |

**1502 of 1531 candidates (98.1%)** carried a job record under a sibling agent.
The count was predicted from a job-record scan *before* the import ran and
matched the measured result exactly.

Controlled pair, content held constant:
- transcript with a sibling-agent job record → offered, "adopted", **invisible**
- byte-identical copy under a fresh session id, no job record → offered, adopted,
  **visible**

The UI states the contradiction plainly: a toast reading *"Imported 96 chats"*
while the chat count moves 36 → 38.

Genuine terminal `claude` history (no job records) imports correctly — this only
bites where the source directory also holds another agent's sessions. But an
embedding app that points an agent at its own working directory hits it every
time.

## Suggested fix

Exclude anything already attributed to **any** agent, not merely anything whose
origin isn't native:

```js
if (attribution.origin !== "native" || attribution.agentName !== undefined) {
  skipped.push({ sessionId, reason: "attributed-to-run", detail: ... });
  continue;
}
```

Worth considering alongside it:

- **Make the two gates share one predicate.** They encode the same question in
  two places and drifted; a single `isAdoptableUnderAgent()` used by both would
  make this structurally impossible rather than fixed once.
- **Defensive check at adoption time**, not just at scan time — `adoptSessionsFrom`
  should refuse to record an adoption that cannot win the precedence contest,
  rather than writing a record that is inert by construction. That is what makes
  the current failure unrecoverable.
- A regression test with a **sibling-agent job record**, since a same-agent record
  behaves differently and would pass a naive test.

## Notes

Found by QA against a real ~1 GB corpus; the synthetic fixtures used during
development had no job records at all, which is why every earlier test passed.

edspencer:
Fixes #437.

## The bug

`scanAdoptionCandidates` asked "is this session **native**?" when it needed to ask "is this session **unattributed**?".

`triggerTypeToOrigin` folds `manual`, `webhook`, `chat` and `fork` into `origin: "native"`. So a session with a genuine job record under a *sibling* agent resolves to `{ origin: "native", agentName: "sweeper-x" }` — natively originated, but firmly someone else's. It passed the adoption gate (`origin !== "native"`) and then failed the per-agent listing gate (`attribution.agentName !== agentName`).

The adoption record was written and lost the precedence contest (job → platform → adopted → native), so it was **inert by construction**. The import reported success, the session never appeared, and the burnt marker read as `already-adopted` on every retry — silent and unrecoverable. Measured on a real corpus: 1502 of 1531 candidates (98.1%) carried a sibling-agent job record.

## What changed

**1. The scan gate.** `isUnattributed` instead of `origin === "native"`. Sessions the *adopting* agent already owns are excluded too — they are already visible under it, so there is nothing to import.

**2. One shared predicate.** The two gates encoded the same question in two places and drifted; that drift *is* the bug. Both now derive from a single set of pure predicates in `session-attribution.ts` — the module that already owns the precedence order — and, crucially, the adoption gate is *defined in terms of* the listing gate:

```ts
canAgentAdopt(attr, agent) === isOwnedByAgent(attributionAfterAdoptionBy(attr, agent), agent)
```

There is one `isOwnedByAgent`, and everything else derives from it, so this class of bug is structurally impossible rather than fixed once. Two invariants follow, and both are asserted directly over the full space of attribution shapes:

- anything the scan offers, the adopt path accepts (`isUnattributed ⇒ canAgentAdopt`)
- anything the adopt path accepts is visible to the adopting agent afterwards

`canAgentAdopt` is deliberately *looser* than `isUnattributed`: re-claiming a session you already own is a harmless no-op, and a session another agent merely *adopted* is adoptable because `recordAdoption` overwrites — adoption records don't outrank a later adoption the way job and platform records do.

**3. Refusal at the write site, not just the scan.** `adoptSession` throws the new `SessionAdoptionRefusedError`; `adoptSessionsFrom` re-checks each candidate immediately before recording it and reports a skip. The scan and the write are separate calls — separated by minutes of file copying in a 1500-transcript import, well past the attribution cache TTL — so a sibling's run can land in between. Refusing is what turns an unrecoverable silent failure into a legible one.

**4. `AdoptSkippedSession.ownedBy`** (new, optional) names the owning agent on `attributed-to-run` and `already-adopted` skips.

### On the third suggestion — splitting `"attributed-to-run"`

I chose **not** to add a skip-reason member, and want to flag it for override.

Every non-native origin also carries an `agentName`, so a new `"attributed-to-other-agent"` member would fire on essentially every real case and leave `"attributed-to-run"` unreachable dead code. Meanwhile adding a union member is the one change here that could break a TS consumer switching exhaustively on the reason — including Paddock, the consumer this unblocks.

An optional `ownedBy` field gives a UI strictly more than a second enum member would (`skip.ownedBy === thisAgent ? "already yours" : \`belongs to ${skip.ownedBy}\``), costs no compatibility, and `detail` was already documented as carrying "the owning agent". If you'd rather have the enum split, it's a small follow-up.

## Tests — red first

Every existing test passed because the #423 fixtures had **no job records at all**. These use real transcripts, real `job-*.yaml` records, the real adoption store and the real attribution index.

Against the unfixed code (with only the additive scaffolding in place), **12 of 16 failed**:

```
listAdoptableSessions > does not offer a session owned by a sibling agent
  AssertionError: expected [ 'sess-sibling' ] to deeply equal []
listAdoptableSessions > ...with trigger_type manual (folds to origin native)
  AssertionError: expected [ 'sess-manual' ] to deeply equal []
listAdoptableSessions > ...with trigger_type webhook (folds to origin native)
  AssertionError: expected [ 'sess-webhook' ] to deeply equal []
listAdoptableSessions > ...with trigger_type chat (folds to origin native)
  AssertionError: expected [ 'sess-chat' ] to deeply equal []
listAdoptableSessions > ...with trigger_type fork (folds to origin native)
  AssertionError: expected [ 'sess-fork' ] to deeply equal []
listAdoptableSessions > still offers a genuinely unattributed transcript alongside an owned one
  AssertionError: expected [ 'sess-owned', 'sess-free' ] to deeply equal [ 'sess-free' ]
listAdoptableSessions > does not offer a session the adopting agent already owns via its own job record
  AssertionError: expected [ 'sess-mine' ] to deeply equal []
adoptSessionsFrom > skips a sibling-owned session and names the owner
  AssertionError: expected [ 'sess-sibling' ] to deeply equal []
adoptSessionsFrom > reports exactly the sessions that then become visible
  AssertionError: expected [ 'sess-free-a', 'sess-free-b', …(3) ] to deeply equal [ 'sess-free-a', 'sess-free-b' ]
adoptSessionsFrom > refuses a candidate that becomes sibling-owned between the scan and the write
  AssertionError: expected [ 'sess-b', 'sess-a' ] to deeply equal [ 'sess-a' ]
adoptSession > refuses a session that became sibling-owned between the scan and the adopt
  AssertionError: promise resolved "{ version: 1, …(4) }" instead of rejecting
adoptSession > carries the owning agent on the refusal
  AssertionError: expected null to be an instance of SessionAdoptionRefusedError
```

`reports exactly the sessions that then become visible` is the headline symptom in miniature: the unfixed code adopts 5 and shows 2.

The **4 that passed** against the unfixed code are the ones that matter for interpreting the rest:

- `control: a "web" job record was already excluded before the fix` — kept deliberately. `trigger_type: "web"` maps to origin `"web"`, so the old gate already caught it. A regression test built on that trigger type proves nothing; this one documents why the four native-folding types are the whole bug.
- the three `adoptSession` "still allows…" cases (unattributed, same-agent, handover-from-another-adopter) — pinning that the new refusal isn't over-broad.

The same-agent / sibling-agent distinction is pinned on both paths, since a same-agent record behaves differently and would pass a naive test.

## Verification

- `packages/core`: **1 failed | 3661 passed | 1 skipped (3663)**. The single failure is the known pre-existing `directory.test.ts > "throws StateDirectoryCreateError when parent directory is not writable"`, which fails only because this box runs tests as root. It is the only one.
- chat 298, cli 202 (+10 skipped), web 176, discord 342 (+97 skipped), slack 350 — all passing.
- `tsc --noEmit` clean across all packages; `biome check` clean on every touched file; `pnpm --filter @herdctl/core build` succeeds and the module loads.
- The new test file was run 3× to confirm it isn't order-dependent (candidate ordering is pinned with explicit mtimes).

## Notes for review

- **Behaviour change:** `adoptSession` now throws where it previously returned an inert record, and `listAdoptableSessions` no longer offers sessions the agent already owns via a job record. Adopting genuine terminal `claude` history — the case adoption exists for — is unaffected.
- The write-site re-check calls `getAttributionIndex()` per candidate. That's TTL-cached, so it's a map lookup in the common case and rebuilds only once the TTL lapses mid-batch, which is exactly the case it's there to catch.
- Three existing test files mocked `session-attribution.js` wholesale; they now spread the actual module (the pattern already used for the `session-metadata` mock) so the pure predicates aren't stubbed out.
- Changeset included as a `patch`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Session adoption now targets only truly unowned sessions.
  - Prevented conflicting adoptions when another agent already owns a session.
  - Added safeguards against ownership changes during adoption.
  - Sessions already owned by the current agent are excluded from adoption listings.
  - Batch adoption results now identify skipped sessions and their current owner.

- **Documentation**
  - Clarified session ownership, adoption eligibility, refusal behavior, and skip details.

### DIFF
--- a/.changeset/lucky-poems-fetch.md
+++ b/.changeset/lucky-poems-fetch.md
@@ -1,0 +1,50 @@
+---
+"@herdctl/core": patch
+---
+
+Fix session adoption offering — and silently losing — sessions owned by another agent (#437)
+
+`scanAdoptionCandidates` asked "is this session **native**?" when it needed to
+ask "is this session **unattributed**?". `triggerTypeToOrigin` folds `manual`,
+`webhook`, `chat` and `fork` into `origin: "native"`, so a session with a real
+job record under a *sibling* agent resolved to
+`{ origin: "native", agentName: "sweeper-x" }` — natively originated, but firmly
+someone else's. It passed the adoption gate and then failed the per-agent
+listing gate (`agentName !== agentName`).
+
+The result was an adoption record that could never win the precedence contest
+(job → platform → adopted → native): inert by construction. The import reported
+success, the session never appeared, and the burnt marker read as
+`already-adopted` on every retry — silent and unrecoverable. Measured on a real
+corpus, 1502 of 1531 candidates (98.1%) carried a sibling-agent job record; one
+workspace reported "imported 1382" and 26 appeared. Any app that points several
+agents at one working directory hits this every time.
+
+- **Only unattributed sessions are adoptable.** `listAdoptableSessions` and
+  `adoptSessionsFrom` now skip anything an agent already owns — including the
+  adopting agent itself, whose sessions are already visible and so have nothing
+  to import.
+- **One shared predicate.** The listing gate and the adoption gate encoded the
+  same question in two places and drifted; that drift *was* the bug. Both now
+  derive from `isOwnedByAgent` in `session-attribution.ts`, with the adopt-time
+  gate defined as "would the listing gate pass afterwards"
+  (`canAgentAdopt = isOwnedByAgent(attributionAfterAdoptionBy(...))`). Two
+  invariants follow and are covered by tests: anything the scan offers, the
+  adopt path accepts; anything the adopt path accepts is then visible.
+- **Refusal at the write site, not just the scan.** `adoptSession` throws the
+  new `SessionAdoptionRefusedError`, and `adoptSessionsFrom` re-checks each
+  candidate immediately before recording it and reports a skip instead. The scan
+  and the write are separate calls — separated by minutes of file copying in a
+  large import, well past the attribution cache TTL — so a sibling's run can
+  land in between. Turning that into a legible skip is what keeps it
+  recoverable.
+- **`AdoptSkippedSession.ownedBy`** (new, optional) names the owning agent on
+  `attributed-to-run` and `already-adopted` skips, so a UI can distinguish
+  "belongs to another agent" from "already yours". Deliberately a field rather
+  than a new `AdoptSkipReason` member, which would break consumers that switch
+  on the reason exhaustively.
+
+Behaviour change worth noting: `adoptSession` now throws where it previously
+returned an inert record, and `listAdoptableSessions` no longer offers sessions
+the agent already owns via a job record. Adopting genuine terminal `claude`
+history — the case adoption exists for — is unaffected.
--- a/docs/src/content/docs/architecture/session-discovery.md
+++ b/docs/src/content/docs/architecture/session-discovery.md
@@ -360,8 +360,8 @@ const discovery = new SessionDiscoveryService({
 | `getSessionMetadata(workDir, sessionId)` | Get metadata for a session (cached) |
 | `getSessionUsage(workDir, sessionId)` | Get token usage data for a session |
 | `invalidateCache(workDir?)` | Clear cached data for a specific directory or all caches |
-| `listAdoptableSessions(agentName, workDir, fromWorkingDir?)` | List native transcripts an agent could adopt (see [Session Adoption](#session-adoption)) |
-| `adoptSession(agentName, sessionId, opts)` | Record an adoption claim for a single session, moving nothing |
+| `listAdoptableSessions(agentName, workDir, fromWorkingDir?)` | List unattributed transcripts an agent could adopt (see [Session Adoption](#session-adoption)) |
+| `adoptSession(agentName, sessionId, opts)` | Record an adoption claim for a single session, moving nothing. Refuses if another agent owns it |
 | `adoptSessionsFrom(agentName, workDir, opts?)` | Place and claim every adoptable session found in a directory |
 | `unadoptSession(sessionId, opts?)` | Remove an adoption record, leaving the transcript on disk |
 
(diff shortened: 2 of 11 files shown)


edspencer:
Declining the one actionable review finding, with reasoning.

**The suggestion:** `FleetManager.adoptSession` does not catch `SessionAdoptionRefusedError`, so it propagates; catch it and report the owning agent instead.

**Why not.** The two adoption entry points are deliberately different shapes, and both are already correct:

- **`adoptSessionsFrom` (batch)** — the path every UI actually uses — never throws. It checks attribution up front and records a skip via `describeAttributionSkip`, and its per-session `catch` maps any late failure to `record-failed`. One bad session cannot abort the batch. Verified: paddock's `POST …/adopt-chats` calls `adoptSessionsFrom`, not `adoptSession`, so the live path is unaffected by this finding.
- **`adoptSession` (single)** — a caller naming one specific session and asking for it to be adopted. If another agent owns it, that request cannot be satisfied, and a typed, documented error is the honest answer. It already carries the owning agent name and the detail string, so nothing is lost by throwing.

Converting that throw into a quiet return value would make it easy to ignore — which is precisely the failure mode this PR exists to eliminate. #437 was an unrecoverable silent no-op: adoption reported success while writing a record that was inert by construction. Trading a loud typed failure for a soft one re-creates that class of bug one layer up.

Embedding apps that want soft behaviour already have it: use the batch API, which reports refusals as skips rather than raising.

PR #441, opened 2026-08-04
feat(core): reapChatSession — close a managed session the reap policy never releases

edspencer:
## The gap

`decideReap` keeps a streaming session alive for exactly as long as it holds live background work, and says so plainly in its own header: *"No idle timer, no max-lifetime backstop, no idle-concurrency cap."*

That is the right default — reaping a session with work in flight kills the work. But it left consumers with **no way out** when a session becomes permanently unreapable, and there are at least two ordinary ways that happens:

1. **A background task never exits.** A model-authored `until` loop whose sentinel never arrives, so `backgroundTasks` never drains.
2. **A re-invocation turn dies without firing a Stop hook** — on a subscription usage limit, say. `activity` has already cleared `awaitingTasks`, so every later `background_tasks_changed` returns early at the `isAwaitingTasks()` guard, and only a `turn_end` could re-arm it or reap. None comes. The session is stranded live with no pending reap.

In both cases the session's message stream never ends, so a consumer rendering that stream shows the session as running until the process restarts. There was no API to end it.

## The change

- **`SessionReaper.forceReap(sessionId): boolean`** — close a live managed session regardless of what it is holding.
- **`FleetManager.reapChatSession(sessionId): boolean`** — public exposure, since `sessionLifecycle` is private on the fleet.

Both idempotent; both return `false` for an unknown, unmanaged or already-reaped id.

## Why on the reaper, and not just `close()`

A consumer holds the `RuntimeSession` and could call `close()` itself. That is **not** equivalent — it closes the query behind the reaper's bookkeeping:

- `liveById` keeps a stale entry, so `whenSessionReaped` never resolves and a later resume of that id stalls until its 5-minute ceiling (#403).
- `WakeRegistry` skips that session's wakes indefinitely.

`forceReap` routes through the same private `reap` the policy uses, so `markDone()` unregisters the id and drains the reap waiters before `close()`. The consumer observes an ordinary end-of-stream and unwinds through its ordinary path — no new teardown contract to get right.

Worth stating explicitly: **`interrupt()` is not a substitute.** It targets an in-flight model turn, and a session held open purely for background work has none.

## Scope

Policy is untouched — nothing reaps on its own that didn't before. This only adds a door that can be opened deliberately.

The reap log line now carries a reason; `Reaping idle session …` is unchanged, a forced one reads `Reaping force-reaped on request session …`.

## Tests

Four new reaper tests and three FleetManager tests. The one that matters is `rescues a session stranded live by a re-invocation turn that never ends` — it drives the real signal sequence (`turn_end` keepAlive → `background_tasks_changed` drained → `activity`, then nothing) and asserts the session is genuinely stuck (`onReap` never fires, `close()` never called) before showing `forceReap` releases it.

`pnpm typecheck` green across the monorepo; core suite passes (3675) and the docs site builds. One pre-existing failure locally, `directory.test.ts > throws StateDirectoryCreateError when parent directory is not writable`, is environmental — it fails on `main` too because this box runs as uid 0, so a chmod-unwritable parent is still writable.

## Docs

`library-reference/fleet-manager.mdx` gains a `reapChatSession()` section; `concepts/sessions.md` notes that the keep-alive rule has no backstop and points at it.

Unblocks edspencer/paddock#528.

## Summary by CodeRabbit

- **New Features**
  - Added an API to force-close live managed chat sessions on demand.
  - Returns whether a session was successfully closed and safely handles unknown or already-closed sessions.
  - Forced closures use standard cleanup and are recorded distinctly from automatic reaping.

- **Documentation**
  - Documented session behavior when background work remains unfinished.
  - Added reference documentation for the new session-reaping API.

- **Tests**
  - Added coverage for forced closure, idempotency, waiting callers, and sessions stranded during unfinished work.

### DIFF
--- a/.changeset/force-reap-session.md
+++ b/.changeset/force-reap-session.md
@@ -1,0 +1,49 @@
+---
+"@herdctl/core": minor
+---
+
+feat(core): `FleetManager.reapChatSession` — close a managed session that the reap policy will never release
+
+`decideReap` keeps a streaming session alive for exactly as long as it holds
+live background work, with — by its own header comment — "no idle timer, no
+max-lifetime backstop, no idle-concurrency cap". That is the right default:
+reaping a session with work in flight kills the work. But it left consumers with
+no way out when a session becomes permanently unreapable, and there are at least
+two ordinary ways that happens:
+
+- A background task never exits (a model-authored `until` loop whose sentinel
+  never arrives), so `backgroundTasks` never drains.
+- A re-invocation turn dies without firing a Stop hook — on a subscription usage
+  limit, say. `activity` has already cleared `awaitingTasks`, so every later
+  `background_tasks_changed` returns early at the guard, and only a `turn_end`
+  could re-arm it or reap. None comes. The session is stranded live with no
+  pending reap.
+
+In both cases the session's message stream never ends, so a consumer rendering
+that stream shows the chat as running until the process restarts. There was no
+API to end it.
+
+**`SessionReaper.forceReap(sessionId)`** closes a live managed session
+regardless of what it is holding, and **`FleetManager.reapChatSession(sessionId)`**
+exposes it (the lifecycle manager is private on the fleet). Both are idempotent
+and return `false` for an unknown, unmanaged or already-reaped id.
+
+This belongs on the reaper rather than being a `close()` the consumer calls on
+the `RuntimeSession` it already holds. Closing the query directly goes behind the
+reaper's bookkeeping: `liveById` keeps a stale entry, so `whenSessionReaped`
+never resolves — a later resume of that id stalls until its 5-minute ceiling
+(#403) — and `WakeRegistry` skips that session's wakes forever. `forceReap`
+routes through the same private `reap` the policy uses, so the id is
+unregistered, reap waiters drain, and the consumer observes an ordinary
+end-of-stream and unwinds through its ordinary path. No new teardown contract.
+
+The reap log line now carries a reason (`Reaping idle session …` is unchanged;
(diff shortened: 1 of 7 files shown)


edspencer:
> **Mock the file system in this test.** Use the repository's in-memory filesystem utility or mock the filesystem calls.

Skipping this one — the premise doesn't hold on two counts:

1. **There is no in-memory filesystem utility in this repo.** No `memfs`, `mock-fs` or `unionfs` in any `package.json`, and no shared test-fs helper. The suggestion points at something that doesn't exist.

2. **Real temp dirs are the established pattern for FleetManager tests** — 14 of the 17 files in `packages/core/src/fleet-manager/__tests__/` use `mkdtemp`. `reap-chat-session.test.ts` is deliberately modelled on `session-control.test.ts`, which does the same mkdtemp/mkdir/writeFile/rm dance for the same reason: `FleetManager.initialize()` reads a real `herdctl.yaml` and initialises a real state dir, and that wiring is exactly what the test exists to exercise. It reaches `sessionLifecycle` — a private field — through the public `getSessionLifecycle()`, so stubbing the fs out would leave it asserting against a construction path that isn't the production one.

The guideline it cites ("mock external dependencies") is aimed at the SDK, network and GitHub API — non-deterministic or costly dependencies. A `mkdtemp` under `os.tmpdir()`, torn down in `afterEach`, isn't in that category, and the policy behaviour itself is covered with no filesystem at all in `session/__tests__/session-reaper.test.ts`.

Happy to revisit if an in-memory fs helper is introduced and the other 14 files move over — but this one test shouldn't be the odd one out.

PR #446, opened 2026-08-05
feat(core): plugins passthrough and full MCP server config (#444, #445)

### LINKED ISSUE #444: No plugins passthrough: SDK-run agents cannot load Claude Code plugins
Found while wiring host MCP-server inheritance in Paddock (edspencer/paddock#691 / #699).

## Summary

An agent run through herdctl cannot use the operator's installed Claude Code **plugins**, and there is no configuration that makes it possible. Two independent blockers, both in herdctl rather than the SDK:

1. **No passthrough.** The Agent SDK accepts `plugins?: SdkPluginConfig[]` (`{type: 'local', path}`). `SDKQueryOptions` has no such field and `toSDKOptions` never sets one — `plugins` appears exactly once in `dist/`, in a comment. So an embedder has no channel to write to.

2. **Auto-discovery is gated behind a setting herdctl excludes.** The SDK *does* auto-discover plugins from `join(CLAUDE_CONFIG_DIR, "plugins")` and merges them with anything passed via the option — verified with a live probe against a planted home. But discovery is gated on `enabledPlugins`, which lives in `settings.json`'s **user** source, and herdctl invokes the SDK with `--setting-sources=project` for every agent with a working directory. The flag is therefore never read, so the discovered plugins stay disabled.

Worth stating because the obvious diagnosis is wrong: it is not that the SDK ignores installed plugins. It discovers them and then never enables them.

## Impact

Plugins provide commands, agents, skills, hooks **and MCP servers**. A user with, say, a Slack plugin providing an MCP server sees it work in their terminal `claude` and silently not exist under herdctl — no error, no log line.

## Possible fixes, roughly in order of size

- **Add a `plugins` passthrough** to `SDKQueryOptions` / agent config, plumbed through `toSDKOptions`. Smallest, and enough for an embedder that enumerates plugin paths itself.
- **Allow `settingSources` to be configured** per agent (or include `user` when the embedder asks). Would make auto-discovery work as the CLI does, but pulls in the rest of user settings, which some embedders deliberately exclude — Paddock now has explicit `instructions` / `hooks` levers precisely to control that, so a blunt `user` would fight it.
- Both: passthrough for explicit control, `settingSources` for parity with the CLI.

Not urgent for us — Paddock's #699 ships the MCP half and documents the plugin half as blocked rather than working around it. Filing so the constraint is recorded upstream rather than rediscovered.

### LINKED ISSUE #445: McpServerSchema drops headers and type, silently breaking authenticated and SSE MCP servers
Found while wiring host MCP-server inheritance in Paddock (edspencer/paddock#691 / #699).

## Summary

`McpServerSchema` is `{command?, args?, env?, url?}` — a plain `z.object`, so every other key is stripped at `addAgent`. Two of the stripped keys are load-bearing for real-world servers:

| dropped | consequence |
|---|---|
| `headers` | bearer / API-key auth is lost, so an authenticated remote server fails to connect |
| `type: "sse"` | `transformMcpServer` maps every `url` to `type: "http"`, so an SSE server is misconfigured |

Both are accepted by the Agent SDK's own `McpServerConfig` (`McpSSEServerConfig`, `McpHttpServerConfig`), so this is herdctl narrowing a shape the layer beneath it supports.

## The part that makes it worse than a dropped field

Claude Code keys a remote server's stored OAuth token on a hash of the server config:

```
`${name}|${sha256({type, url, headers}).slice(0, 16)}`
```

So dropping `headers` or rewriting `type` does not merely lose auth on the wire — it **changes the key the token is filed under**, and a previously-authorised server is no longer recognised as the same server. The user re-authorises, and it happens again next run.

## Suggested fix

Widen `McpServerSchema` to match the SDK's `McpServerConfig`: keep `headers`, and honour an explicit `type` (`"sse"` / `"http"`) rather than inferring `http` from the presence of `url`. `transformMcpServer` then only needs to default `type` when absent.

## Workaround in place downstream

Paddock (#699) passes such servers anyway and emits a `warn` naming each affected server at boot, on the grounds that a capability lever should degrade loudly rather than silently. It cannot fix the hash problem from outside.

Low urgency — stdio servers (the common case) are unaffected. Filing so it is recorded rather than rediscovered by whoever next tries an authenticated remote server.

edspencer:
Fixes #444. Fixes #445.

Both issues are the same defect class — herdctl narrowing what the Agent SDK supports underneath it, silently — and they touch adjacent code, so they ship together.

## What changed

**#445 — `McpServerSchema` dropped `headers` and `type`.** The schema was `{command, args, env, url}`, a plain `z.object`, so every other key was stripped at `addAgent`; `transformMcpServer` then rewrote every `url` to `type: "http"`. It now also accepts `headers`, an explicit `type` (`"stdio" | "sse" | "http"`), `timeout` and `alwaysLoad`, mirroring the SDK's own `McpServerConfig`. An explicit `type` wins; a bare `url` still infers `http`, so existing configs are untouched.

**#444 — no plugins passthrough.** New optional `plugins` field on an agent config and on fleet `defaults`, plumbed through `toSDKOptions` to the SDK's `plugins` option. Entries are a bare path string or `{type: "local", path, skipMcpDiscovery?}`; the shorthand normalises to the object form.

Both runtimes honour `plugins`: the SDK runtime sets the option, and the CLI runtime emits `--plugin-dir` / `--plugin-dir-no-mcp` per entry — which is exactly what the SDK does with its own option, verified by reading the SDK's bundle. Leaving the CLI runtime out would have reproduced the same silent-narrowing bug one runtime over.

## Design notes for the downstream consumer

Paddock reaches both levers from ordinary agent config, so [paddock#700](https://github.com/edspencer/paddock/issues/700) is unblocked as written:

- `agent.mcp_servers[name]` now survives verbatim, so the `claude.mcpServers: host` merge can stop stripping and stop warning.
- `agent.plugins` is a plain array on the same agent config object `addAgent` already takes.

MCP server field names deliberately mirror the SDK's `McpServerConfig` rather than herdctl's usual snake_case (hence camelCase `alwaysLoad`). An entry is passed through verbatim, so an operator — or Paddock reading `~/.claude.json` — can hand over a block unchanged. A translation layer is exactly where #445's fields got lost.

**I did not widen the default `setting_sources`, and I'd argue against it.** It would fix plugin enablement with no passthrough at all, but it pulls in every other user-source key — user `CLAUDE.md`, `agents/`, `commands/`, `settings.json` hooks — which is the boundary Paddock's `claude.instructions` / `claude.hooks` levers exist to hold closed. A blanket widening would silently undo that. The explicit `plugins` list needs no such grant.

## One correction to #444

> **Allow `settingSources` to be configured** per agent (or include `user` when the embedder asks).

This already exists and already accepts `"user"` — `setting_sources` on `AgentConfigSchema`, honoured by `toSDKOptions` ahead of the `["project"]` default. #444's second blocker therefore already had an operator-facing lever before this PR; what was missing was only the passthrough and the documentation of the trade-off. There is a test in this PR asserting that behaviour, and it is one of the three (of twelve) that already passed on `main`.

Everything else in both issues held up. In particular the non-obvious claim in #444 — that the SDK *discovers* plugins and never *enables* them, rather than ignoring them — is consistent with what the SDK bundle does: `plugins` becomes a `--plugin-dir` CLI flag directly, independent of `enabledPlugins`, which is why the passthrough works without any settings-source change.

## Tests

The important one is `packages/core/src/runner/__tests__/mcp-and-plugin-passthrough.test.ts`. It drives the **whole trip** — `FleetManager.addAgent()` (the real `AgentConfigSchema.parse`) → `getAgents()` → `toSDKOptions()` → the options object actually handed to the SDK's `query()`, with `query` mocked to capture it. That framing is the point: the stripping happened at the config boundary, one layer *above* where the existing `transformMcpServer` unit tests all sat happily passing. **9 of its 12 cases fail on `main`**; the 3 that pass are the deliberate no-regression guards (legacy `url` → `http` inference, the no-plugins case, and the `setting_sources` behaviour noted above).

Also added:
- `sdk-type-contract.test.ts` — compile-time assertions pinning the hand-written `SDKQueryOptions` / `SDKMcpServerConfig` / `SDKPluginConfig` mirrors against the SDK's own types. `sdk-runtime.ts` widens them with `as Record<string, unknown>` before calling `query()`, so nothing else in the build would notice the SDK drifting — which is how `type?: "http"` stayed wrong after the SDK grew `sse`. Reverting the union to `"http"` reproduces `error TS2344: Type '"sse"' does not satisfy the constraint '"http"'`.
- CLI-runtime tests for `--plugin-dir` / `--plugin-dir-no-mcp` and for `headers` + `type: sse` reaching `--mcp-config`.
- Schema tests for the new fields and both plugin entry forms; a `mergeAgentConfig` test for `defaults.plugins` (that function is a per-field allowlist, not a deep merge, so a new field is silently dropped from defaults unless wired in explicitly — it is).

### What the tests do *not* prove

They prove the config survives to the SDK/CLI invocation boundary. They do **not** prove a planted plugin directory actually loads — that needs a real `claude` run against the API, which I did not do. The strongest evidence short of that is from reading the SDK bundle directly: `plugins` is translated to `--plugin-dir <path>` (or `--plugin-dir-no-mcp` when `skipMcpDiscovery` is set), and a non-`local` `type` throws. So the shape is right and reaches the CLI; whether the CLI then loads it is the SDK's contract, not herdctl's.

## Verification

`pnpm typecheck` (11/11), `pnpm build` and `pnpm lint` all pass. `pnpm test` passes except one pre-existing failure unrelated to this branch — `state-manager` "throws StateDirectoryCreateError when parent directory is not writable", which fails on my box because it runs as uid 0, where `chmod 0555` does not prevent writes. It fails identically on unmodified `main`; CI runs non-root, so it should be green here.

## Not done

- No web UI surface for `plugins` (the dashboard does not expose `mcp_servers` either).
- `McpServerSchema` stays permissive — it does not enforce "`url` xor `command`", or that `headers` is only meaningful for a remote server. Adding refinements could reject configs that parse today, and that is a separate call.
- Plugin `path` is used as given, not resolved relative to the agent config file. For a dockerized agent it must resolve inside the container; documented rather than solved.
- `tools` (`McpServerToolPolicy[]`) is still not in `McpServerSchema` — a nested union, not asked for, and no consumer for it yet.

## Release

**Yes — Paddock needs a release to consume this.** There is a changeset (`@herdctl/core`, `minor`), so merging produces a Version Packages PR, and `@herdctl/core` has to be published and Paddock's dependency bumped before paddock#700 can be built against it.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring and passing through local plugins for agents and fleet defaults.
  * Added plugin options to control MCP discovery behavior.
  * Expanded MCP server configuration with HTTP, SSE, and stdio transports, authentication headers, timeouts, and deferred tool loading.
  * Preserved compatibility with existing bare-URL HTTP and stdio configurations.

* **Documentation**
  * Updated configuration and runner guides with plugin and MCP setup details, including authenticated SSE examples.


edspencer:
Thanks — both CodeRabbit findings are declined, with reasons. Summarising here so the decision is on the record rather than buried in a resolved thread.

### 1. "Enforce the required MCP transport fields" — declined

The suggestion is to `refine` `McpServerSchema` so `type: "stdio"` requires `command` and `type: "sse"|"http"` requires `url`.

The problem is **what layer that schema validates**. `McpServerSchema` parses an agent's own config *fragment*, at `AgentConfigSchema.parse()` time — which happens **before** `mergeAgentConfig` folds in fleet `defaults.mcp_servers`. `mcp_servers` is deep-merged, so a partial per-server override is a supported pattern, and the fragment is legitimately incomplete on its own.

I checked this rather than assuming it. Both of these work today and would start throwing at parse time under the proposed refine:

```ts
// Agent overrides only the transport; url comes from fleet defaults.
AgentConfigSchema.parse({ name: "a", mcp_servers: { linear: { type: "sse" } } })
// → merged with defaults {type: "http", url: "https://mcp.linear.app/sse"}
// → {type: "sse", url: "https://mcp.linear.app/sse"}   ✅ valid only after merge
```

```ts
// Agent adds only auth to a server the fleet defines.
AgentConfigSchema.parse({ name: "a", mcp_servers: { linear: { headers: { Authorization: "Bearer x" } } } })
```

A cross-field requirement is meaningful on the *merged* config, not on a fragment. Enforcing it here would reject valid fleet configurations, so if this belongs anywhere it is a post-merge validation pass — out of scope for this PR, and worth its own issue. The PR body already lists the permissive schema as a deliberate non-goal; this is the concrete reason.

### 2. "Reject `headers` on cleartext MCP URLs" — declined

Requiring `https://` whenever `headers` is non-empty would break the most common authenticated-MCP setup there is: a server on loopback. herdctl itself already emits exactly that shape — `packages/core/src/runner/runtime/cli-runtime.ts:298` injects `http://127.0.0.1:${bridge.port}/mcp` for its own MCP HTTP bridge. Auth headers against `127.0.0.1` are not cleartext transmission over a network in any meaningful sense.

More importantly, `.mcp.json` and the Agent SDK both accept `headers` with an `http://` URL. Rejecting one here would mean a config that works in `claude` fails under herdctl — **which is precisely the defect class this PR exists to fix** (#444/#445 are both "herdctl narrowing a shape the layer beneath it supports"). Adding a new narrowing while removing two would be self-defeating.

Worth noting the finding's own reachability path lists `packages/core/src/config/__tests__/merge.test.ts` as the external entry point, which is a test file, not an external input.

If cleartext-plus-auth is worth surfacing, the right shape is a `warn` at boot naming the server — loud, not fatal — and that is a separate change from this one.

PR #450, opened 2026-08-12
feat(core): stop a single background task in a live session (#449)

### LINKED ISSUE #449: Expose stopTask on RuntimeSession, and a session-id-keyed door that outlives the turn
## Problem

A consumer cannot stop **one** background task in a live session. Two separate
gaps, and both have to close before it works.

Context: the Agent SDK exposes `Query.stopTask(taskId)` — a `stop_task` control
request. It is already available in the SDK version `@herdctl/core` has installed
(`0.3.215`, `sdk.d.ts:2535`), so no dependency bump is needed.

### Gap 1 — `RuntimeSession` does not forward it

`packages/core/src/runner/runtime/interface.ts` — `RuntimeSession` has six
members: `messages`, `send`, `interrupt`, `listCommands`, `setModel`, `close`.

The `Query` itself is unreachable. In `sdk-runtime.ts` `SDKRuntime.openSession`,
`q` is a `const` captured by the closures of the returned object literal; it is
never assigned to `this`, never returned, and there is no generic
control-request passthrough.

*(One apparent escape hatch is not one: `messages` is `q` behind a cast, so at
runtime `session.messages` **is** the `Query`. But any consumer passing
`onLifecycleSignal` — which is the interesting case — gets the
`tapLifecycleStream` wrapper generator instead, which forwards nothing.)*

### Gap 2 — the handle does not outlive the turn, but the tasks do

This is the one that actually bites, and it is why this should be solved here
rather than in each consumer.

A consumer holds a `RuntimeSession` for the duration of a turn and releases it
when the turn's result lands. Background tasks — background shells, sub-agents,
monitors — routinely **outlive that turn**. So even with Gap 1 closed, "stop this
task" only works while a turn happens to be in flight, and goes inert exactly
when it is most wanted: the long tail where something is still running and the
user is looking at it wondering why.

Consumers cannot fix this by holding the handle longer, because they do not own
the session's lifetime — the reaper does. And herdctl **already** keeps the live
session past the turn: `SessionReaper.liveById: Map<string, ManagedSessionImpl>`
(`session-reaper.ts:115`), with `ManagedSessionImpl` holding
`private readonly session: RuntimeSession` (`:356`). The state is here; there is
just no door onto it.

## Proposed API

**1. `RuntimeSession.stopTask`** — `runner/runtime/interface.ts`:

```ts
/** Stop one background task by id. The SDK emits a terminal
 *  `task_notification{status:"stopped"}` on the message stream. */
stopTask(taskId: string): Promise<void>;
```

and in `sdk-runtime.ts` `openSession`'s returned literal, beside `setModel`:

```ts
stopTask: (taskId: string) => q.stopTask(taskId),
```

**2. A session-id-keyed door** — mirroring the `reapChatSession` / `forceReap`
pair that already exists for the same reason:

```ts
// SessionReaper
stopTaskInSession(sessionId: string, taskId: string): Promise<boolean>;

// FleetManager, beside reapChatSession (fleet-manager.ts:1168)
stopTaskInSession(sessionId: string, taskId: string): Promise<boolean>;
```

`false` when the session is not live — same contract shape as `reapChatSession`.

`openSession` is optional on `RuntimeInterface` and only `SDKRuntime` implements
it, so this is SDK-runtime-only by construction. Decide explicitly what the
session-id path does when the live session is a runtime without `stopTask`.

## Semantics worth preserving (from the CLI's own handler)

- **Idempotent.** `not_found` / `not_running` are converted to a *success*
  response, so stopping something that just finished is not an error. Do not add
  a liveness pre-check that reintroduces the race.
- **Not ownership-gated.** The in-session `TaskStop` tool checks the caller owns
  the task; the control-request path passes no caller id, so an out-of-band actor
  may stop any task in the session. That is what makes this useful.
- **Self-notifying.** The kill emits the terminal `task_notification` itself, so
  consumers need no separate confirmation and no bookkeeping.
- **`monitor_mcp` has no kill strategy** in the CLI and **rejects** with
  `unsupported_type` (unlike `not_found`, this one is not swallowed).
  Recommendation: let the rejection through rather than translating it to a
  silent success — a consumer needs to know the button did nothing.

## Notes for implementation

- A changeset; minor on `@herdctl/core`.
- Test doubles that build fake `RuntimeSession` objects will need the new method:
  `session-lifecycle-manager.test.ts`, `session-reaper.test.ts`,
  `open-chat-session-resume-collision.test.ts`, `reap-chat-session.test.ts`.
- `RuntimeSession` is already re-exported through `runner/runtime/index.ts` →
  `runner/index.ts` → `src/index.ts`, so a new method rides along with no barrel
  change. `job-control.ts` returns the runtime object verbatim.
- Encouragingly, core already parses background-task state
  (`session-hooks.ts` `toBackgroundTaskSummaries`, `reaper-policy.ts` `keepAlive`
  on `background_tasks`) — task ids already flow through here.

## Consumer

Paddock [#848](https://github.com/edspencer/paddock/issues/848) — a per-row stop
button on its running-work bar. It already holds `(sessionId → taskId)` in
memory; it is blocked on exactly these two gaps, and on Gap 2 specifically for
the background phase.

edspencer:
Closes #449.

Two gaps stood between a consumer and stopping **one** background task in a live session. Both are closed here.

## Gap 1 — `RuntimeSession.stopTask(taskId)`

`SDKRuntime.openSession` captured the SDK `Query` as a `const` in a closure: never assigned to `this`, never returned, no generic control-request passthrough. `session.messages` *is* the `Query` behind a cast, but any caller passing `onLifecycleSignal` — the interesting case, and the one the reaper uses — gets the `tapLifecycleStream` wrapper generator instead, which forwards nothing. So the method is added to the interface and forwarded in the returned literal beside `setModel`. No dependency bump: `stopTask` is already in the installed SDK (`0.3.215`).

## Gap 2 — `FleetManager.stopTaskInSession(sessionId, taskId)`

The half that actually bites. Background shells, sub-agents and monitors routinely outlive the turn that started them, while a consumer holds its `RuntimeSession` only until the turn's result lands. A handle-scoped stop therefore works while a turn happens to be in flight and goes inert exactly over the long tail where something is still running and the user is asking why. Consumers can't fix that by holding the handle longer — they don't own the session's lifetime, the reaper does, and `SessionReaper.liveById` already keeps the session across that gap. This is the door onto that state, mirroring the existing `reapChatSession` / `forceReap` pair: `SessionReaper.stopTaskInSession` under `FleetManager.stopTaskInSession`.

Semantics carried over from the CLI's own handler: **idempotent** (`not_found` / `not_running` are successes, and no liveness pre-check is added that would reintroduce that race), **not ownership-gated** (the control request carries no caller id — that's what makes an out-of-band stop button work), and **self-notifying** (the SDK emits the terminal `task_notification{status:"stopped"}` on the session's own stream).

## The three decisions

**1. A live session whose runtime cannot stop tasks → throw, don't return `false`.**

Moot today — `stopTask` is a *required* member of `RuntimeSession`, and `openSession` is optional on `RuntimeInterface` with only `SDKRuntime` implementing it — but the handle is a plain structural object supplied by whoever opened the session, so an older build, a foreign runtime or a stale test double can satisfy the type nominally and still not carry the method. Resolving quietly there would report a still-running task as stopped, which is the one outcome a stop button must never produce. So it throws a new `SessionTaskControlUnsupportedError` (extends `FleetManagerError`, code `SESSION_TASK_CONTROL_UNSUPPORTED`, exported from the barrel). The payoff is that `false` keeps a single unambiguous meaning: *no live session with this id*.

**2. `monitor_mcp`'s `unsupported_type` rejection propagates.** Agreed with your recommendation. A `monitor_mcp` task has no kill strategy in the CLI, and translating its rejection into a success would render a stopped-looking row for a task that is still running. Same reasoning as decision 1, so the two are consistent: every "the stop did not happen" outcome surfaces as a rejection, and `false` never means failure. A rejection is also inert with respect to reaper state — the session stays live, registered, and reapable — which the tests pin.

**3. `false` (not throw) for a session that is not live.** Matches `reapChatSession`. "The session is already gone" is a race a consumer will hit routinely — a reap landing between render and click — and it isn't exceptional: the tasks went with it, so the user's intent is already satisfied. `Promise<boolean>` rather than `boolean` since the underlying control request is async.

## Tests

- `runner/runtime/__tests__/session-stop-task.test.ts` — the delegation reaches `Query.stopTask` with the right id; still reaches it when a lifecycle tap has replaced the message stream (the case that closes the `messages`-as-escape-hatch loophole); stops nothing else (no `interrupt`, no `return`); a refusal propagates.
- `session/__tests__/stop-task-in-session.test.ts` — resolves a live session by id; `false` for unknown / reaped / detached; addresses the right session when several are live; a refusal leaves the session live, registered and still reapable; the unsupported-handle throw; and no liveness pre-check on the task id.
- `fleet-manager/__tests__/stop-task-in-session.test.ts` — drives the **real** `SessionLifecycleManager` + `SessionReaper` the fleet builds in `initialize()`, as `reap-chat-session.test.ts` does. Includes the Gap-2 case explicitly: a turn runs, ends leaving live background work, and the task is then stopped by session id alone — the `RuntimeSession` handle is only ever read as a spy from that point, never called through. It asserts the session survives the stop (not closed, not interrupted), so it can't pass against a change that only fixes Gap 1.

The four existing fakes named in the issue gained the new member.

## Verification

`pnpm typecheck` green across all 7 packages, `pnpm build` green (including docs), biome clean on the changed files. Full suite: 3727 passed, 1 pre-existing failure unrelated to this change — `state/__tests__/directory.test.ts > throws StateDirectoryCreateError when parent directory is not writable`, which fails identically on a clean `main` on this box because it runs as uid 0 and `chmod 0o555` doesn't stop root. Core coverage 85.25% lines / 77.41% branches, above the package's 74/65 gates.

Changeset: minor on `@herdctl/core`. Docs: a `stopTaskInSession` section in the FleetManager reference plus a pointer from the sessions concept page.

## Consumer

Unblocks edspencer/paddock#848 — a per-row stop button on its running-work bar, which needs Gap 2 specifically for the background phase.

## Summary by CodeRabbit

* **New Features**
  * Added the ability to stop individual background tasks in active sessions without ending the session or interrupting other tasks.
  * Added session-level task stopping by session ID, including idempotent handling of completed tasks.
  * Added clear errors when task control is unavailable and surfaced runtime refusal errors.
  * Added notifications when tasks are stopped.

* **Documentation**
  * Documented task stopping and its behavior for active, unavailable, and already-closed sessions.

### DIFF
--- a/.changeset/stop-task-in-session.md
+++ b/.changeset/stop-task-in-session.md
@@ -1,0 +1,41 @@
+---
+"@herdctl/core": minor
+---
+
+feat(core): stop a single background task in a live session
+
+A consumer could not stop **one** background task in a session — two gaps, both
+of which had to close.
+
+**`RuntimeSession.stopTask(taskId)`** forwards the SDK's `stop_task` control
+request. The SDK `Query` was a `const` captured in `SDKRuntime.openSession`'s
+closure: never assigned to `this`, never returned, with no generic passthrough,
+so `stopTask` was unreachable from outside. (`session.messages` is the `Query`
+behind a cast, but any caller passing `onLifecycleSignal` — the interesting
+case — gets a wrapper generator that forwards nothing.)
+
+**`FleetManager.stopTaskInSession(sessionId, taskId)`** (and
+`SessionReaper.stopTaskInSession` beneath it) addresses that stop by session id
+instead of by handle. That is the half that actually bites: background shells,
+subagents and monitors routinely outlive the turn that started them, while a
+consumer holds its `RuntimeSession` only until the turn's result lands. A
+handle-scoped stop therefore works only while a turn happens to be in flight and
+goes inert over exactly the long tail where something is still running and the
+user is asking why. Consumers cannot hold the handle longer, because they do not
+own the session's lifetime — the reaper does, and already keeps the session past
+the turn. This is the door onto that state, mirroring `reapChatSession` /
+`forceReap`.
+
+Semantics carried over from the CLI's own handler: idempotent (`not_found` /
+`not_running` are successes, so stopping a task that just finished is fine — and
+no liveness pre-check is added that would reintroduce that race); not
+ownership-gated, so an out-of-band actor such as a UI stop button can stop any
+task in the session; and self-notifying, since the SDK emits the terminal
+`task_notification{status:"stopped"}` on the session's own stream.
+
+`stopTaskInSession` returns `false` only for a session that is not live, matching
+`reapChatSession`. A runtime refusal propagates rather than being translated into
+a silent success — notably `monitor_mcp` tasks, which the CLI has no kill
+strategy for and rejects with `unsupported_type`; a consumer needs to know the
+button did nothing. A live session whose handle carries no `stopTask` at all
+throws the new `SessionTaskControlUnsupportedError` for the same reason.
--- a/docs/src/content/docs/concepts/sessions.md
+++ b/docs/src/content/docs/concepts/sessions.md
@@ -240,6 +240,8 @@ A live streaming session keeps a warm `claude` process around (~300 MB each). Se
 
 The keep-alive rule has no backstop — no idle timer, no max lifetime. That is deliberate (reaping a session with work in flight kills the work), but it means a session whose background work never finishes is never reaped on its own, and its message stream never ends. [`reapChatSession()`](/library-reference/fleet-manager/#reapchatsessionsessionid) closes such a session on demand, so a UI can offer a working "stop".
 
+For the narrower case — one runaway background task, not the whole session — [`stopTaskInSession()`](/library-reference/fleet-manager/#stoptaskinsessionsessionid-taskid) stops a single shell, subagent or monitor by id and leaves the session running. Both are keyed by session id for the same reason: background work outlives the turn that started it, so the `RuntimeSession` handle a consumer holds is already gone by the time the stop is wanted.
+
 Wake semantics:
 
 | Behavior | Rule |
--- a/docs/src/content/docs/library-reference/fleet-manager.mdx
+++ b/docs/src/content/docs/library-reference/fleet-manager.mdx
@@ -1486,6 +1486,56 @@ if (!manager.reapChatSession(sessionId)) {
 
 ---
 
+### `stopTaskInSession(sessionId, taskId)`
+
+Stops one background task in a managed chat session, leaving the session running.
+
+```typescript
+await manager.stopTaskInSession(
+  sessionId: string,
+  taskId: string
+): Promise<boolean>
+```
+
+#### Parameters
+
+| Parameter | Type | Required | Description |
+|-----------|------|----------|-------------|
+| `sessionId` | `string` | **Yes** | Resolved session id of a session opened with `manageLifecycle: true` |
+| `taskId` | `string` | **Yes** | Id of the background task, as carried on the session's `task_notification` messages and on the `BackgroundTaskSummary` handed to `onKeepAlive` |
+
+#### Returns
+
+`true` once the stop request has been delivered; `false` for an unknown, unmanaged or already-reaped session id. Rejects if the runtime refuses the stop — see below.
+
+#### Description
+
+The finer-grained sibling of [`reapChatSession()`](#reapchatsessionsessionid): that ends the whole session, this ends a single background shell, sub-agent or monitor and leaves everything else running. The SDK emits the terminal `task_notification` with `status: "stopped"` on the session's own message stream, so a consumer driving a UI off that stream needs no separate confirmation and no bookkeeping.
+
+Why by session id rather than on the `RuntimeSession` you hold? Because background tasks routinely outlive the turn that started them, and your handle does not — you release it when the turn's result lands. A stop reachable only through the handle would work while a turn happened to be in flight and go inert over exactly the window where a task is still running and the user is asking why. Past the turn, the reaper is the only thing still holding the session, so that is where the door has to be.
+
+<Aside type="note">
+`interrupt()` is not a substitute here either. It targets the in-flight model turn; during the background phase there is none, and interrupting would not touch the task anyway.
+</Aside>
+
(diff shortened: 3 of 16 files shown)


edspencer:
CI is green (the one red Unit Tests run was `scheduler.test.ts:418` "triggers cron schedules when time is within trigger window" — a wall-clock flake in code this PR doesn't touch; all 51 scheduler tests pass locally and the job passes on re-run).

On CodeRabbit's three comments — skipping all three, reasons below. Happy to be overruled on any of them.

**1. `await` in the docs signature block.** Technically right that it isn't valid TypeScript, but every async method in `fleet-manager.mdx` is documented this way (`await manager.initialize(): Promise<void>`, `await manager.stop(...): Promise<void>`, `await manager.listAgentCommands(...): Promise<SlashCommand[]>`, …). Diverging for this one section would read as the inconsistency. Worth a separate sweep across the file if we want to change the convention.

**2. Mock the filesystem in `stop-task-in-session.test.ts`.** This would remove the thing the test is for. It's modelled on `reap-chat-session.test.ts`, which uses a real tempdir deliberately so the test drives the *real* `SessionLifecycleManager` + `SessionReaper` that `initialize()` builds — that's the wiring under test, reached through a public method rather than a stub. Mocking the FS turns it into an assertion about a mock. The reaper-level contract is already covered without any filesystem at `session/__tests__/stop-task-in-session.test.ts`. `mkdtemp` is also the established pattern here: 30 occurrences across `fleet-manager/__tests__/`.

**3. Zod-validate `sessionId` / `taskId`.** No public `FleetManager` method validates its string params this way — `fleet-manager.ts` has zero Zod usage, including the sibling `reapChatSession(sessionId)`. The convention targets *external* input (config files, CLI args, API payloads), not library method arguments the type system already covers. There's also no unsafe path to guard: a non-string `sessionId` misses the `liveById` lookup and returns `false`, which is the documented answer for "not live". Adding validation to this one method would be an inconsistency without a payoff.

